PR #18, opened 2017-09-28
Scenarios and jobs run and wait

nicolasservel:


dataikuapi/dss/project.py
```diff
@@ -276,6 +279,26 @@ def start_job(self, definition):
...
+        job_def = self.client._perform_json("POST", "/projects/%s/jobs/" % self.project_key, body = definition)
+        job = DSSJob(self.client, self.project_key, job_def['id'])
+        job_state = job.get_status().get("baseStatus", {}).get("state", "")
+        sleep_time = 2
+        while job_state not in ["DONE", "ABORTED", "FAILED"]:
+            sleep_time = 300 if sleep_time >= 300 else sleep_time * 2
+            time.sleep(sleep_time)
+            job_state = job.get_status().get("baseStatus", {}).get("state", "")
+            if job_state in ["ABORTED", "FAILED"]:
+                raise DataikuException("Job run did not finish. Status: %s" % (job_state))
```
FChataigner: rather have 2 modes, controlled by parameter to the func (like "no_fail=False") 
- if it fails raise exception, else return
- never raise exception but return final status

=> to avoid having to try: ... except: ... around a call of start_job_and_wait()

PR #29, opened 2018-10-18
Adding saved model delete

alexbourret:


dataikuapi/dss/savedmodel.py
```diff
@@ -109,3 +109,14 @@ def get_object_discussions(self):
         :rtype: :class:`dataikuapi.discussion.DSSObjectDiscussions`
         """
         return DSSObjectDiscussions(self.client, self.project_key, "SAVED_MODEL", self.sm_id)
+
+    ########################################################
+    # Deletion
+    ########################################################
+
+    def delete(self, drop_data=False):
```
cstenac:
```suggestion
    def delete(self):
```

Unused argument

PR #32, opened 2018-10-30
Prediction type option in ML task creation

LouisPlt:
Add optional parameters to force prediction type in ML pred task crea……tion, and add a method to re-guess the task parameters with a option to force pred type


LouisPlt:
[ch28316](https://app.clubhouse.io/dataiku/story/28316/ability-to-force-prediction-type-when-creating-ml-task-using-api)

PR #39, opened 2019-01-31
Add available methods to a DSS project. 

wnowak10:
Couldn't access these methods using autocomplete so I explicitly added here to doc strings.

(Please let me know if this method (branch 5.1, commit, PR back to 5.1) is inappropriate.


AlexReutter:
I'm not certain how to test this without a dev install, which I don't currently have.  @apichery is there someone you can recommend to review this change?

cstenac:
Class doc shouldn't be a substitute for autocompletion or autodoc

PR #44, opened 2019-10-09
Client WebApp API

MayeulRousselet:
Implementation of a WebApp API client + TU


spicquenot (changes requested):
Refactorize the WebApp classes for simplicity (see DSSWiki & DSSWikiSettings as example)

dataikuapi/dss/webapp.py
```diff
@@ -0,0 +1,90 @@
...
+    A handle to manage a webapp
+    """
+    def __init__(self, client, project_key, webapp_id, state=None):
+        """Do not call directly, use :meth:`dataikuapi.dss.project.DSSProject.get_webapps`"""
+        self.client = client
+        self.project_key = project_key
+        self.webapp_id = webapp_id
+        self.state = state
+
+    def get_state(self):
```
FChataigner: make a call to fetch state if self.state is None. We can't demand that you only use DSSWebapp objects coming from the list() call

and the state is important, people will use it to monitor what is running or not...

FChataigner (changes requested):

spicquenot (changes requested):
small reqs

dataikuapi/dss/webapp.py
```diff
@@ -0,0 +1,128 @@
...
+        self.state = None
+
+    def get_state(self):
+        """
+        Return the state of the webapp
+
+        :return: the state of the webapp
+        :rtype: :class:`DSSWebAppBackendState`
+        """
+        if self.state is None:
```
spicquenot: I would not store/cache the state of the backend, otherwise, the user cannot get a refreshed state for one instance of a DSSWebApp (calling get_state several times will always result of the same state)

dataikuapi/dss/webapp.py
```diff
@@ -0,0 +1,128 @@
...
+        return "futureInfo" in self.definition and \
+               "alive" in self.definition["futureInfo"] and \
+               self.definition["futureInfo"]["alive"]
+
+
+class DSSWebAppDefinition(object):
+    """
+    A handle to manage a WebApp definition
+    """
+    def __init__(self, webapp, client, project_key, webapp_id, definition):
```
spicquenot: if we remove storing the state, we can remove the webapp reference here

dataikuapi/dss/webapp.py
```diff
@@ -0,0 +1,128 @@
...
+        """
+        definition = self.client._perform_json("GET", "/projects/%s/webapps/%s" % (self.project_key, self.webapp_id))
+        return DSSWebAppDefinition(self, self.client, self.project_key, self.webapp_id, definition)
+
+
+class DSSWebAppBackendState(object):
+    """
+    A handle to manage WebApp backend state
+    """
+    def __init__(self, client, project_key, webapp_id, definition):
```
spicquenot: I would rename definition to state to avoid confusion (not stable definition here, but changing state of the backend)

dataikuapi/dss/webapp.py
```diff
@@ -0,0 +1,128 @@
...
+        """Do not call directly, use :meth:`dataikuapi.dss.webapp.DSSWebApp.get_state`"""
+        self.client = client
+        self.project_key = project_key
+        self.webapp_id = webapp_id
+        self.definition = definition
+
+    def get_definition(self):
```
spicquenot: ditto get_state

dataikuapi/dss/webapp.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def save(self):
+        """
+        Save the current webapp definition and update it.
+
+        :rtype: :class:`dataikuapi.dss.future.DSSFuture`
+        """
+        self.client._perform_json("PUT", "/projects/%s/webapps/%s" % (self.project_key, self.webapp_id), body=self.definition)
+        self.definition = self.client._perform_json("GET", "/projects/%s/webapps/%s" % (self.project_key, self.webapp_id))
+        self.webapp.state = None
```
spicquenot: ditto remove

spicquenot (changes requested):
should not store the state within the webapp, otherwise, cannot refresh the state, even by calling get_state on the webapp object

spicquenot (changes requested):
better future handling

spicquenot (approved):
all good, if you want you can add a new function in DSSWebAppBackendState get_future that returns a DSSFuture with the jobId stored in DSSWebAppBackendState so the loading state of the webapp backend can be handle more easily

FChataigner (approved):
LGTM

tests/webapps_tests.py
```diff
@@ -0,0 +1,82 @@
+from time import sleep
```
apichery, line 1: I would remove this test class as it cannot be run automatically

apichery (changes requested):

dataikuapi/dss/webapp.py
```diff
@@ -0,0 +1,121 @@
...
+        :rtype: :class:`DSSWebAppBackendState`
+        """
+        state = self.client._perform_json("GET", "/projects/%s/webapps/%s/backend/state" % (self.project_key, self.webapp_id))
+        return DSSWebAppBackendState(self.client, self.project_key, self.webapp_id, state)
+
+    def stop_backend(self):
+        """
+        Stop a webapp
+        """
+        self.client._perform_empty("PUT", "/projects/%s/webapps/%s/backend/actions/stop" % (self.project_key, self.webapp_id))
```
cstenac, line 36: Check with @FChataigner whether it's a good thing that this does not return a future

dataikuapi/dss/webapp.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def restart_backend(self):
+        """
+        Restart a webapp
+        :returns: a handle to a DSS future to track the progress of the restart
+        :rtype: :class:`dataikuapi.dss.future.DSSFuture`
+        """
+        future = self.client._perform_json("PUT", "/projects/%s/webapps/%s/backend/actions/restart" % (self.project_key, self.webapp_id))
+        return DSSFuture(self.client, future["jobId"])
```
cstenac, line 45: use DSSFuture.from_resp which properly handles the case where the future instantly succeeded or failed (here, it will do KeyError on jobId

dataikuapi/dss/webapp.py
```diff
@@ -0,0 +1,121 @@
...
+        state = self.client._perform_json("GET", "/projects/%s/webapps/%s/backend/state" % (self.project_key, self.webapp_id))
+        return DSSWebAppBackendState(self.client, self.project_key, self.webapp_id, state)
+
+    def stop_backend(self):
+        """
+        Stop a webapp
+        """
+        self.client._perform_empty("PUT", "/projects/%s/webapps/%s/backend/actions/stop" % (self.project_key, self.webapp_id))
+
+    def restart_backend(self):
```
cstenac, line 38: While this corresponds to what the API really does, the fact that there is a "restart" method but no "start" method could be confusing. 

Not sure what the best to do is here. Maybe call the method `start_or_restart_backend`

dataikuapi/dss/webapp.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        Restart a webapp
+        :returns: a handle to a DSS future to track the progress of the restart
+        :rtype: :class:`dataikuapi.dss.future.DSSFuture`
+        """
+        future = self.client._perform_json("PUT", "/projects/%s/webapps/%s/backend/actions/restart" % (self.project_key, self.webapp_id))
+        return DSSFuture(self.client, future["jobId"])
+
+    def get_definition(self):
```
cstenac, line 47: "get_definition" is the legacy naming for legacy methods that return a dict. Modern methods that return settings must be called get_settings and return an object called XXXSettings

dataikuapi/dss/webapp.py
```diff
@@ -0,0 +1,121 @@
...
+    A handle to manage WebApp backend state
+    """
+    def __init__(self, client, project_key, webapp_id, state):
+        """Do not call directly, use :meth:`dataikuapi.dss.webapp.DSSWebApp.get_state`"""
+        self.client = client
+        self.project_key = project_key
+        self.webapp_id = webapp_id
+        self.state = state
+
+    def get_state(self):
```
cstenac, line 69: Make this a property

dataikuapi/dss/webapp.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        Returns the dict containing the current state of the webapp backend.
+        Warning : this dict is replaced when webapp backend state changes
+
+        :returns: a dict
+        """
+        return self.state
+
+    def is_running(self):
```
cstenac, line 78: Make this a property called "running"

dataikuapi/dss/webapp.py
```diff
@@ -0,0 +1,121 @@
...
+    A handle to manage a WebApp definition
+    """
+    def __init__(self, client, project_key, webapp_id, definition):
+        """Do not call directly, use :meth:`dataikuapi.dss.webapp.DSSWebApp.get_definition`"""
+        self.client = client
+        self.project_key = project_key
+        self.webapp_id = webapp_id
+        self.definition = definition
+
+    def get_definition(self):
```
cstenac, line 100: So this would be webapp.get_definition().get_definition() :)  Standard practice is to call this method get_raw

so it will be webapp.get_settings().get_raw()

dataikuapi/dss/webapp.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        Get the definition
+
+        :returns: the definition of the webapp
+        """
+        return self.definition
+
+    def set_definition(self, definition):
```
cstenac, line 108: No, no set function on a "modern" settings object. It's modification in place.

tests/webapps_tests.py
cstenac: Talk with @Basharsh96 first :)

cstenac (changes requested):
API modernization

tests/webapps_tests.py
Basharsh96: I made a copy of the test and I'll add it later to the tests in the CI pipeline I'm working on. Thank you, you can safely remove it 😊

PR #48, opened 2020-01-03
Public API for manipulating statistics worksheets

fterrazzoni:


dataikuapi/dss/worksheet.py
```diff
@@ -0,0 +1,120 @@
...
+
+        Returns:
+            the definition, as a JSON object
+        """
+        return self.client._perform_json(
+            "GET", "/projects/%s/statistics/worksheets/%s" % (self.project_key, self.worksheet_id))
+
+    def set_definition(self, definition):
+        """
+        Set the definition of the dataset
```
instanceofme:
```suggestion
        Set the definition of the worksheet
```

dataikuapi/dss/worksheet.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def __init__(self, client, project_key, data_spec, card):
+        self.client = client
+        self.project_key = project_key
+        self.data_spec = data_spec
+        self.card = card
+
+    def compute(self):
+        """
+        Compute the results of a single card (without worksheet)
```
instanceofme:
```suggestion
        Compute the results of this single card (without worksheet)
```

fterrazzoni:
Here is how the client is supposed to be used 


```python
# Get a dataset
dataset = client.get_project('LambdaBinaryClassifMllibTest').get_dataset('kaggle_titanic_train_copy')

# List associated worksheets
worksheets = dataset.list_statistics_worksheets()

# Create a new worksheet
worksheet = dataset.create_statistics_worksheet()

# Obtain settings
worksheet_settings = worksheet.get_settings()

# Add a card
worksheet_settings.add_card({"type": 'correlation_matrix', "columns": [{"name": "Age", "type": "CONTINUOUS"} , {"name": "Fare", "type": "CONTINUOUS"} ], "heatmapParams": {}})

# Save worksheet
worksheet_settings.save()

# Compute worksheet
worksheet.run_worksheet().wait_for_result()

# List cards 
cards = worksheet_settings.list_cards()

# Compute a card
card = cards[0]
worksheet.run_card(card).wait_for_result()

# Get computation from a card
computation = card.compile()

# Run computation
worksheet.run_computation(computation)

# Run (any) computation with dict
worksheet.run_computation({"type": "count"}).wait_for_result()

# Delete the worksheet
worksheet.delete()
```

fterrazzoni:
[playwithedanotebook.pdf](https://github.com/dataiku/dataiku-api-client-python/files/4200421/playwithedanotebook.pdf)

instanceofme:
- [ ] After talking, we actually want to remove internal dict as its 3 methods are not wanted:
    - `get` is useless
    - `__repr__` exposes inner data, which can be accessed with `get_raw` and for which we want to favor helpers over direct access
    - `get_raw` is nice but its doc should be overridden, so not worth factorizing
- [ ] [Nitpick / opinion] could inline the future wrappers

PR #50, opened 2020-01-24
Implement wrappers for post train computation results

nicolasservel:
First implementation for subpopulation. 

- [x] Still need to implement it for partial dependencies

### DIFF
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -1,6 +1,7 @@
 from ..utils import DataikuException
 from ..utils import DataikuUTF8CSVReader
 from ..utils import DataikuStreamedHttpUTF8CSVReader
+from ..utils import DSSExtendableDict
 import json
 import time
 from .metrics import ComputedMetrics
@@ -606,8 +607,8 @@ class DSSTrainedPredictionModelDetails(DSSTrainedModelDetails):
         :param int n_jobs: number of cores used for parallel training. (-1 means 'all cores')
         :param bool debug_mode: if True, output all logs (slower)
 
-        :returns: if wait is True, a dict containing the Subpopulation analyses, else a future to wait on the result
-        :rtype: dict or :class:`dataikuapi.dss.future.DSSFuture`
+        :returns: if wait is True, an object containing the Subpopulation analyses, else a future to wait on the result
+        :rtype: :class:`dataikuapi.dss.ml.DSSSubpopulationAnalyses` or :class:`dataikuapi.dss.future.DSSFuture`
         """
         
         body = {
@@ -633,26 +634,30 @@ class DSSTrainedPredictionModelDetails(DSSTrainedModelDetails):
             )
             future = DSSFuture(self.saved_model.client, future_response.get("jobId", None), future_response)
         if wait:
-            return future.wait_for_result()
+            return DSSSubpopulationAnalyses(future.wait_for_result())
         else:
             return future
 
 
     def get_subpopulation_analyses(self):
         """
-        Retrieve all subpopulation analyses computed for this trained model as a dict
+        Retrieve all subpopulation analyses computed for this trained model
+        
+        :returns: the subpopulation analyses
+        :rtype: :class:`dataikuapi.dss.ml.DSSSubpopulationAnalyses`
         """
         
         if self.mltask is not None:
-            return self.mltask.client._perform_json(
+            data = self.mltask.client._perform_json(
                 "GET", "/projects/%s/models/lab/%s/%s/models/%s/subpopulation-analyses" %
                 (self.mltask.project_key, self.mltask.analysis_id, self.mltask.mltask_id, self.mltask_model_id)
             )
         else:
-            return self.saved_model.client._perform_json(
+            data = self.saved_model.client._perform_json(
                 "GET", "/projects/%s/savedmodels/%s/versions/%s/subpopulation-analyses" %
                 (self.saved_model.project_key, self.saved_model.sm_id, self.saved_model_version),
             )
+        return DSSSubpopulationAnalyses(data)
 
     def compute_partial_dependencies(self, features, wait=True, sample_size=1000, random_state=1337, n_jobs=1, debug_mode=False):
         """
@@ -711,6 +716,74 @@ class DSSTrainedPredictionModelDetails(DSSTrainedModelDetails):
                 "GET", "/projects/%s/savedmodels/%s/versions/%s/partial-dependencies" %
                 (self.saved_model.project_key, self.saved_model.sm_id, self.saved_model_version),
             )
+
+class DSSSubpopulationAnalysis(DSSExtendableDict):
+    """
+    Object to read details of a subpopulation analysis of a trained model
+
+    Do not create this object directly, use :meth:`DSSSubpopulationAnalyses.get_analysis(feature)` instead
+    """
+
+    def __init__(self, analysis):
+        super(DSSSubpopulationAnalysis, self).__init__(analysis)
+
+    def get_computation_params(self):
+        """
+        Gets computation params
+        """
+        computation_params = {}
+        computation_params["nbRecords"] = self.get("nbRecords")
+        computation_params["randomState"] = self.get("randomState")
+        computation_params["onSample"] = self.get("onSample")
+        return computation_params
+
+    def get_raw(self):
+        """
+        Gets the raw dictionary of the subpopulation analysis
+        """
+        return self.internal_dict
+
+
+class DSSSubpopulationAnalyses(DSSExtendableDict):
+    """
+    Object to read details of subpopulation analyses of a trained model
+
+    Do not create this object directly, use :meth:`DSSTrainedPredictionModelDetails.get_subpopulation_analyses()` instead
+    """
+
+    def __init__(self, data):
+        super(DSSSubpopulationAnalyses, self).__init__(data)
+        self.analyses = []
+        for analysis in data.get("subpopulationAnalyses", []):
+            self.analyses.append(DSSSubpopulationAnalysis(analysis))
+    
+    def get_raw(self):
+        """
+        Gets the raw dictionary of subpopulation analyses
+        """
+        return self.internal_dict
+    
+    def get_all_dataset(self):
+        """
+        Retrieve information and performance on the full dataset used to compute the subpopulation analyses
+        """
+        return self.get("allDataset")
+
+    def list_analyses(self):
+        """
+        Lists all features on which subpopulation analyses have been computed
+        """
+        return [analysis["feature"] for analysis in self.analyses]
+    
+    def get_analysis(self, feature):
+        """
+        Retrieves the subpopulation analysis for a particular feature
+        """
+        if feature not in self.list_analyses():
+            raise ValueError("Subpopulation analysis for feature '%s' cannot be found" % feature)
+
+        return next(analysis for analysis in self.analyses if analysis["feature"] == feature)
+
 
 class DSSClustersFacts(object):
     def __init__(self, clusters_facts):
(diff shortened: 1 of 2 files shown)


dataikuapi/utils.py
```diff
@@ -94,3 +94,65 @@ def str_to_bool(s):
                                                 doublequote=True):
                 yield [none_if_throws(caster)(val)
                         for (caster, val) in dku_zip_longest(casters, uncasted_tuple)]
+
+class DSSExtendableDict(dict):
```
instanceofme: - I think the correct word is `Extensible`
- Could use a docstring specifying in which way this is or isn't compatible with `dict`

fterrazzoni:
Out of curiosity, is it a new pattern we plan to generalize to other APIs ? (ie. replace get_definition() by inheriting a dict-like object)

cc @instanceofme

dataikuapi/dss/ml.py
```diff
@@ -722,6 +722,86 @@ def get_partial_dependencies(self):
...
+    def get_definition(self):
+        """
+        Gets the definition of the subpopulation analysis modality
+
+        :returns: definition
+        :rtype: :class:`dataikuapi.dss.ml.DSSSubpopulationModalityDefinition`
+        """
+        return self.definition
+    
+    def excluded(self):
```
instanceofme: Consider a clearer version:

```suggestion
    def is_excluded(self):
```

How is it done elsewhere?

dataikuapi/dss/ml.py
```diff
@@ -722,6 +722,86 @@ def get_partial_dependencies(self):
...
+        return self.get("perf")
+
+
+class DSSSubpopulationModalityDefinition(object):
+
+    def __init__(self, data):
+        self.missing_values = data.get("missing_values")
+        self.index = data.get("index")
+    
+    def missing_values(self):
```
instanceofme: Consider clearer version:

```suggestion
    def is_missing_values(self):
```

How is it done elsewhere?

dataikuapi/dss/ml.py
```diff
@@ -741,6 +823,44 @@ def get_computation_params(self):
...
+    def get_modality(self, definition=None, missing_values=False):
+        """
+        Retrieves modality from definition
+
+        :param definition: definition of modality to retrieve. Can be:
+                   * :class:`dataikuapi.dss.ml.DSSSubpopulationModalityDefinition`
+                   * for category modality, can be a str corresponding to the value of the modality
+                   * for numeric modality, can be a number inside the modality
+        :param missing_values: whether to retrieve modality corresponding to missing values. If True, 
+                               `definition` is ignored
```
instanceofme: Alternative option to consider: replace `missing_value` argument by a call with `definition= DSSSubpopulationModalityDefinition.MISSING_VALUES` (being a constant Definition)

dataikuapi/dss/ml.py
```diff
@@ -741,6 +823,44 @@ def get_computation_params(self):
...
+        if isinstance(definition, DSSSubpopulationModalityDefinition):
+            modality_candidates = [m for m in self.modalities if m.definition.index == definition.index]
+            if len(modality_candidates) == 0:
+                raise ValueError("Modality with index '%s' not found" % modality["index"])
+            return modality_candidates[0]
+        
+        for m in self.modalities:
+            if m.definition.contains(definition):
+                return m
+        raise ValueError("Modality not found")
```
instanceofme:
```suggestion
        raise ValueError("Modality not found: %s" % m)
```

dataikuapi/dss/ml.py
```diff
@@ -692,25 +697,280 @@ def compute_partial_dependencies(self, features, wait=True, sample_size=1000, ra
...
+
+    def get_computation_params(self):
+        """
+        Gets computation params
+        """
+        computation_params = {}
+        computation_params["nbRecords"] = self.get("nbRecords")
+        computation_params["randomState"] = self.get("randomState")
+        computation_params["onSample"] = self.get("onSample")
+        return computation_params
```
instanceofme: [Nitpick] Why not direct literal?

```python
return {
    "nbRecords": self.get("nbRecords"),
    "randomState": self.get("randomState"),
    "onSample": self.get("onSample")
}
```

dataikuapi/dss/ml.py
```diff
@@ -692,25 +697,280 @@ def compute_partial_dependencies(self, features, wait=True, sample_size=1000, ra
...
+        return [partial_dep["feature"] for partial_dep in self.partial_dependencies]
+
+    def get_partial_dependence(self, feature):
+        """
+        Retrieves the subpopulation analysis for a particular feature
+        """
+        if feature not in self.list_partial_dependencies():
+            raise ValueError("Partial dependence for feature '%s' cannot be found" % feature)
+
+        return next(pd for pd in self.partial_dependencies if pd["feature"] == feature)
```
instanceofme: [Nitpick] I'd write it as catch:

```python
try:
    return next(pd for pd in self.partial_dependencies if pd["feature"] == feature)
except StopIteration:
    raise ValueError("Partial dependence for feature '%s' cannot be found" % feature)
```

dataikuapi/utils.py
```diff
@@ -94,3 +94,65 @@ def str_to_bool(s):
...
+    def popitem(self):
+        return self.internal_dict.popitem()
+
+    def pop(self, key, *argv):
+        return self.internal_dict.pop(key, *argv)
+
+    def setdefault(self, key, default_value=None):
+        return self.internal_dict.setdefault(key, default_value)
+
+    def update(self,*args, **kwargs):
```
instanceofme:
```suggestion
    def update(self, *args, **kwargs):
```

dataikuapi/dss/ml.py
nicolasservel: Yeah that's better, thx

dataikuapi/dss/ml.py
```diff
@@ -633,26 +634,30 @@ def compute_subpopulation_analyses(self, split_by, wait=True, sample_size=1000,
...
             )
         else:
-            return self.saved_model.client._perform_json(
+            data = self.saved_model.client._perform_json(
                 "GET", "/projects/%s/savedmodels/%s/versions/%s/subpopulation-analyses" %
                 (self.saved_model.project_key, self.saved_model.sm_id, self.saved_model_version),
             )
+        return DSSSubpopulationAnalyses(data)
 
     def compute_partial_dependencies(self, features, wait=True, sample_size=1000, random_state=1337, n_jobs=1, debug_mode=False):
```
instanceofme: - Auto-wrap `features` if a single string is provided, otherwise the error message is hard to understand (cannot parse the class from body)
- Ditto for `compute_subpopulation_analyses`, auto-wrap `split_by`

dataikuapi/dss/ml.py
```diff
@@ -692,25 +697,282 @@ def compute_partial_dependencies(self, features, wait=True, sample_size=1000, ra
...
+            "onSample":  self.get("onSample")
+        }
+    
+    def list_modalities(self):
+        """
+        List definitions of modalities
+        """
+        return [m.definition for m in self.modalities]
+
+    def get_modality(self, definition=None):
```
instanceofme: Rename

```suggestion
    def get_modality_data(self, definition=None):
```

dataikuapi/dss/ml.py
```diff
@@ -692,25 +697,282 @@ def compute_partial_dependencies(self, features, wait=True, sample_size=1000, ra
...
+        for analysis in data.get("subpopulationAnalyses", []):
+            self.analyses.append(DSSSubpopulationAnalysis(analysis))
+    
+    def get_raw(self):
+        """
+        Gets the raw dictionary of subpopulation analyses
+        """
+        return self.internal_dict
+    
+    def get_all_dataset(self):
```
instanceofme: Not the usual "dataset", and may make it seem like it's the full unsampled dataset. Rename:

```suggestion
    def get_global(self):
```

dataikuapi/dss/ml.py
```diff
@@ -692,25 +697,282 @@ def compute_partial_dependencies(self, features, wait=True, sample_size=1000, ra
...
+        for pd in data.get("partialDependencies", []):
+            self.partial_dependencies.append(DSSPartialDependence(pd))
+
+    def get_raw(self):
+        """
+        Gets the raw dictionary of partial dependencies
+        """
+        return self.internal_dict
+
+    def list_partial_dependencies(self):
```
instanceofme: Rename:

```suggestion
    def list_features(self):
```

dataikuapi/utils.py
```diff
@@ -94,3 +94,74 @@ def str_to_bool(s):
...
+        return self.internal_dict[key]
+
+    def __iter__(self):
+        return self.internal_dict.__iter__()
+
+    def __setitem__(self, key, value):
+        self.internal_dict[key] = value
+
+    def __str__(self):
+        return self.internal_dict.__str__()
```
instanceofme: - To make it clear (when exploring in a notebook) that it's not actually a dict and that it also has some additional methods, we may want to do something like `FinalClassName({internal_dict_str})`
- Not sure if we also want to override `__repr__`?

dataikuapi/dss/ml.py
nicolasservel: In that case, we should probably rename the field in the data structure as well, it's decided in memory, so no need for migration

instanceofme: Sure

dataikuapi/utils.py
nicolasservel: So, after some [reading](https://www.journaldev.com/22460/python-str-repr-functions), no need to implement `__str__` if `__repr__` returns a string, so only implement `__repr__`. Thx for pointing this out

dataikuapi/dss/ml.py
```diff
@@ -692,25 +697,282 @@ def compute_partial_dependencies(self, features, wait=True, sample_size=1000, ra
...
         else:
-            return self.saved_model.client._perform_json(
+            data = self.saved_model.client._perform_json(
                 "GET", "/projects/%s/savedmodels/%s/versions/%s/partial-dependencies" %
                 (self.saved_model.project_key, self.saved_model.sm_id, self.saved_model_version),
             )
+        return DSSPartialDependencies(data)
+
+
+class DSSSubpopulationModality(DSSExtensibleDict):
```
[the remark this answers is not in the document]
nicolasservel: You are talking about `DSSSubpopulationModalityDefinition` ? If yes, I did it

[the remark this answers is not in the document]
instanceofme: Yes

dataikuapi/dss/ml.py
```diff
@@ -692,25 +697,282 @@ def compute_partial_dependencies(self, features, wait=True, sample_size=1000, ra
...
+        """
+        return self.definition
+    
+    def is_excluded(self):
+        """
+        Whether modality has been excluded from analysis (e.g. too few rows in the subpopulation)
+        """
+        return self.get("excluded", False)
+
+    def get_perf(self):
```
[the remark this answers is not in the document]
nicolasservel: In order no to duplicate logic to build performance metrics that is already done in the backend, adding `performanceMetrics` field in the subpop results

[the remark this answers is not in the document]
instanceofme: Once that PR is merged, we can resolve this conversation

dataikuapi/dss/ml.py
```diff
@@ -723,6 +723,44 @@ def get_partial_dependencies(self):
         return DSSPartialDependencies(data)
 
 
+class DSSSubpopulationGlobal(DSSExtensibleDict):
+    """
+    Object to read details of performance on global dataset used for subpopulation analyses.
```
instanceofme: Trying to avoid "dataset", don't hesitate if you have a better wording

```suggestion
    Object to read details of performance on global population in a subpopulation analyses.
```

PR #52, opened 2020-01-31
scenario: add delete method

jfyuen:


instanceofme:
[ch45109](https://app.clubhouse.io/dataiku/story/45109/api-to-delete-scenario-including-python-client)

PR #53, opened 2020-02-06
Use the UUID of the page when saving a wiki article. [ch45366]

pbailly:
## Story
[ch45366](https://app.clubhouse.io/dataiku/story/45366/improve-wiki-python-public-api)

## Scope
When using the Python API to save article on the wiki, it will work only when you retrieve the article using its UUID. If you reach the article using its name, the URL to save the article will not be correct.
This PR fix the generated URL by always using the UUID.

## To ease review

Create a new wiki article, and write "version 1" on its body.
Then execute this script:
```client = dataikuapi.DSSClient("http://localhost:8082", "So0oOO0S3kRt")
project = client.get_project("PROJECTID")
wiki = project.get_wiki()

# Check if legacy is still all right
article = wiki.get_article("1")
article.get_data().get_body()
# Should return u'version 1'
data = article.get_data()
data.set_body("version 2")
data.save()

# This is the invalid usecase
article = wiki.get_article("articleId")
article.get_data().get_body()
# Should return u'version 2'
data = article.get_data()
data.set_body("version 3")
# the next line was broken
data.save()

# You can go to your wiki page and see the update
```


spicquenot:
@pbailly I think this way of doing it does not solve the root issue. We changed articles names being the ID of the article to a numerical ID.
The problem is that the DSSWikiArticle constructor take an article_id in parameter whereas it taked in fact a article_id_or_name.
The logical fix would be to change the article_id to article_id_or_name in the constructor and retrieve the article data in the constructor fo the DSSWikiArticle to get the real ID of the article before setting as the attribute:
> self.article_id = article_id
changing it to the article_id gotten by the API call

spicquenot (approved):
lgtm

PR #54, opened 2020-02-06
doc/project: improve doc on set_variables() behavior in a recipe

jfyuen:


dataikuapi/dss/project.py
```diff
@@ -534,7 +534,10 @@ def get_variables(self):
     def set_variables(self, obj):
         """
         Sets the variables of this project.
-        @param obj: must be a modified version of the object returned by get_variables
+        WARNING: when executed in a python recipe (single run), the changes made by `set_variables` will not be "seen" in the recipe.
+                 Use the internal API instead if this behavior is needed
```
instanceofme:
```suggestion
                 Use the internal API dataiku.get_custom_variables() instead if this behavior is needed
```

dataikuapi/dss/project.py
```diff
@@ -534,7 +534,10 @@ def get_variables(self):
     def set_variables(self, obj):
         """
         Sets the variables of this project.
-        @param obj: must be a modified version of the object returned by get_variables
+        WARNING: when executed in a python recipe (single run), the changes made by `set_variables` will not be "seen" in the recipe.
```
instanceofme:
```suggestion
        WARNING: if executed from a python recipe, the changes made by `set_variables` will not be "seen" in that recipe.
```

PR #55, opened 2020-02-10
utils: replace DSSExtensibleDict with DSSInternalDict

jfyuen:


dataikuapi/utils.py
```diff
@@ -95,73 +95,19 @@ def str_to_bool(s):
...
+class DSSInternalDict(object):
     """
-    Utility to define dict-like objects that can be sub-classed.
-    
-    Behaves like dict for most common operations. In particular, 
-    it is possible to update an instance of `:class:`dataikuapi.dss.ml.DSSExtensibleDict` 
-    with either a dict or another instance of `:class:`dataikuapi.dss.ml.DSSExtensibleDict`.
-    
-    Provides an `internal_dict` dict field that is the actual holder of the data.
+    Class that provides some helpers and an `_internal_dict` dict field that is the actual holder of the data.
```
instanceofme: This reduces the scope and responsibility of that class so I think it's now ok to factorize `get_raw` on this class.

jfyuen: Good idea, nice catch.

PR #57, opened 2020-02-14
style: make all mutable function arguments "None" and init in function body

jfyuen:
Mutable arguments were used as parameter to http requests. So they are "side effect" free wrt the runtime, but it is still better to avoid mutability

### DIFF
--- a/dataikuapi/dss/macro.py
+++ b/dataikuapi/dss/macro.py
@@ -33,7 +33,7 @@ class DSSMacro(object):
         return self.definition
 
 
-    def run(self, params={}, admin_params={}, wait=True):
+    def run(self, params=None, admin_params=None, wait=True):
         """
         Run the macro from the project
 
@@ -43,6 +43,10 @@ class DSSMacro(object):
         :param wait: if True, the call blocks until the run is finished
         :returns: a run identifier to use to abort or retrieve results
         """
+        if params is None:
+            params = {}
+        if admin_params is None:
+            admin_params = {}
         return self.client._perform_json(
             "POST", "/projects/%s/runnables/%s" % (self.project_key, self.runnable_type), 
             params={'wait':wait}, body={'params':params, 'adminParams':admin_params})['runId']
--- a/dataikuapi/dss/macro.py
+++ b/dataikuapi/dss/macro.py
@@ -39,7 +39,7 @@ class DSSMacro(object):
 
         :param dict params: parameters to the macro run (defaults to `{}`)
         :param dict admin_params: admin parameters to the macro run (if the authentication of
-                             the api client does not cover admin rights, they are ignored, default to empty)
+                             the api client does not cover admin rights, they are ignored, default to `{}`)
         :param wait: if True, the call blocks until the run is finished
         :returns: a run identifier to use to abort or retrieve results
         """
--- a/dataikuapi/dss/macro.py
+++ b/dataikuapi/dss/macro.py
@@ -37,9 +37,9 @@ class DSSMacro(object):
         """
         Run the macro from the project
 
-        :param params: parameters to the macro run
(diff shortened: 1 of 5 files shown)


dataikuapi/dssclient.py
```diff
@@ -966,7 +978,7 @@ def execute(self, settings = {}):
         @warning: You must check the 'success' flag
         """
         # Empty JSON dicts can't be parsed properly
-        if settings == {}:
-            settings["_"] = "_"
+        if settings is None:
```
nicolasservel: In that case, if the user actually passes `sesttings={}`, the behaviour changes

jfyuen: Nice catch!

cstenac:
I don't understand what this fixes

instanceofme:
No effective current bug but implements a best practice against Python's surprising behavior of [mutable default arguments](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments):

```python
>>> def func(arg=[]):
...   arg.append(1)
...   return len(arg)
...
>>> func()
1
>>> func()
2
>>> func()
3
```

dataikuapi/dss/project.py
```diff
@@ -676,18 +697,20 @@ def get_scenario(self, scenario_id):
...
         """
         Create a new scenario in the project, and return a handle to interact with it
 
         :param str scenario_name: The name for the new scenario. This does not need to be unique
                                 (although this is strongly recommended)
         :param str type: The type of the scenario. MUst be one of 'step_based' or 'custom_python'
-        :param object definition: the JSON definition of the scenario. Use ``get_definition(with_status=False)`` on an
-                existing ``DSSScenario`` object in order to get a sample definition object
+        :param dict definition: the JSON definition of the scenario. Use ``get_definition(with_status=False)`` on an
+                existing ``DSSScenario`` object in order to get a sample definition object (defaults to `{}`)
```
nicolasservel:
```suggestion
                existing ``DSSScenario`` object in order to get a sample definition object (defaults to `{'params': {}}`)
```

dataikuapi/dss/macro.py
```diff
@@ -33,16 +33,20 @@ def get_definition(self):
...
+    def run(self, params=None, admin_params=None, wait=True):
         """
         Run the macro from the project
 
-        :param params: parameters to the macro run
-        :param admin_params: admin parameters to the macro run (if the authentication of
-                             the api client does not cover admin rights, they are ignored)
+        :param dict params: parameters to the macro run (defaults to `{}`)
+        :param dict admin_params: admin parameters to the macro run (if the authentication of
+                             the api client does not cover admin rights, they are ignored, default to `{}`)
```
nicolasservel:
```suggestion
                             the api client does not cover admin rights, they are ignored, defaults to `{}`)
```

dataikuapi/dssclient.py
```diff
@@ -780,10 +790,15 @@ def get_apideployer(self):
     # Data Catalog
     ########################################################
 
-    def catalog_index_connections(self, connection_names=[], all_connections=False, indexing_mode="FULL"):
+    def catalog_index_connections(self, connection_names=None, all_connections=False, indexing_mode="FULL"):
         """
         Triggers an indexing of multiple connections in the data catalog
+
+        :param list connection_names: list of connections to index, ignored if `all_connections=True` (defaults to `[]`)
+        :param bool all_connections: index all connections, (defaults to `False`)
```
nicolasservel:
```suggestion
        :param bool all_connections: index all connections (defaults to `False`)
```

nicolasservel:
Run locally integration tests (`scenario/ml/prediction/publicapi` and `scenario/misc/publicapi`) and they are all passing, mergin

PR #60, opened 2020-03-03
utils: remove DSSInternalDict

jfyuen:
This PR replaces `DSSInternalDict` (which did not have a lot of features) to use an `_internal_dict` object in each subclass.

### DIFF
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -831,10 +831,7 @@ class DSSSubpopulationGlobal(object):
     """
 
     def __init__(self, data, prediction_type):
-        if data is None:
-            self._internal_dict = dict()
-        else:
-            self._internal_dict = data
+        self._internal_dict = data
         self.prediction_type = prediction_type
 
     def get_raw(self):
@@ -883,11 +880,7 @@ class DSSSubpopulationModality(object):
     """
 
     def __init__(self, feature_name, computed_as_type, data, prediction_type):
-        if data is None:
-            self._internal_dict = dict()
-        else:
-            self._internal_dict = data
-
+        self._internal_dict = data
         self.prediction_type = prediction_type
         if computed_as_type == "CATEGORY":
             self.definition = DSSSubpopulationCategoryModalityDefinition(feature_name, data)
@@ -1101,10 +1094,7 @@ class DSSSubpopulationAnalyses(object):
     """
 
     def __init__(self, data, prediction_type):
-        if data is None:
-            self._internal_dict = dict()
-        else:
-            self._internal_dict = data
+        self._internal_dict = data
         self.prediction_type = prediction_type
         self.analyses = []
         for analysis in data.get("subpopulationAnalyses", []):
@@ -1152,10 +1142,7 @@ class DSSPartialDependence(object):
     """
 
     def __init__(self, data):
-        if data is None:
-            self._internal_dict = dict()
-        else:
-            self._internal_dict = data
+        self._internal_dict = data
 
     def get_raw(self):
         """
@@ -1187,10 +1174,7 @@ class DSSPartialDependencies(object):
     """
 
     def __init__(self, data):
-        if data is None:
-            self._internal_dict = dict()
-        else:
-            self._internal_dict = data
+        self._internal_dict = data
         self.partial_dependencies = []
         for pd in data.get("partialDependencies", []):
             self.partial_dependencies.append(DSSPartialDependence(pd))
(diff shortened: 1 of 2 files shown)


nicolasservel:
Is there any particular rationale ? Because this PR does reintroduce some boiler plate code that was factored in DSSInternalDict

instanceofme:
Yes, the advantages we initially found to `DSSInternalDict` were actually negated:
- The `get` method is not that useful and looks a bit weird
- The `get_raw` would be useful but factorizes the doc whereas we'd like to customize it for each class so we'll have to override anyway
- The `__repr__` exposes internal content whereas we might not want to actively push this when printing the object

All things considered, not that useful ^^

dataikuapi/dss/ml.py
```diff
@@ -824,28 +823,42 @@ def get_partial_dependencies(self):
...
+    def get_raw(self):
+        """
+        Gets the raw dictionary of the actual data
+
+        :rtype: dict
+        """
+        return self._internal_dict
+
+    def __repr__(self):
+        return self.__class__.__name__ + "(" + self._internal_dict.__repr__() + ")"
```
instanceofme: Not sure this is useful

dataikuapi/dss/ml.py
```diff
@@ -862,22 +875,36 @@ def get_prediction_info(self):
...
```
instanceofme: Should return the modality (category value or numerical bin) rather than the full data

dataikuapi/dss/ml.py
```diff
@@ -985,26 +1012,37 @@ def __repr__(self):
...
```
instanceofme: Ditto, can maybe indicate the column name, the type and the number of modalities

dataikuapi/dss/ml.py
```diff
@@ -1046,88 +1084,130 @@ def get_modality_data(self, definition):
...
```
instanceofme: Ditto, maybe just a list of columns for which analyses where computed?

dataikuapi/dss/ml.py
```diff
@@ -1046,88 +1084,130 @@ def get_modality_data(self, definition):
...
```
instanceofme: Ditto → column name + type?

dataikuapi/dss/ml.py
```diff
@@ -1046,88 +1084,130 @@ def get_modality_data(self, definition):
...
```
instanceofme: Ditto → column list?

dataikuapi/dss/ml.py
jfyuen: I couldn't find the type for this class, only the feature.

dataikuapi/dss/ml.py
```diff
@@ -1046,88 +1093,131 @@ def get_modality_data(self, definition):
...
+        Gets the raw dictionary of subpopulation analyses
+
+        :rtype: dict
+        """
+        return self._internal_dict
+
+    def __repr__(self):
+        return "{cls}(prediction_type={type}, analyses={analyses})".format(cls=self.__class__.__name__,
+                                                                           type=self.prediction_type,
+                                                                           analyses=self.list_analyses)
```
nicolasservel:
```suggestion
                                                                           analyses=self.list_analyses())
```

jfyuen:
> I wonder if we need to initialize `_internal_dict` if `data is None` as the class will be totally unusable in that case.

I wondered the same thing. I think it all depends if we want to raise an error on instance creation or when accessing the `dict`. I would prefer to test for `None` at dict creation, but that doesn't prevent using an empty `dict` that would fail later. There is no good way to check everything at object creation...

nicolasservel:
> > I wonder if we need to initialize `_internal_dict` if `data is None` as the class will be totally unusable in that case.
> 
> I wondered the same thing. I think it all depends if we want to raise an error on instance creation or when accessing the `dict`. I would prefer to test for `None` at dict creation, but that doesn't prevent using an empty `dict` that would fail later. There is no good way to check everything at object creation...

I agree with you, there is perfect check. But here, creating an empty dict does not bring anything IMO compared to just initialize the _internal_dict with `None`, it will just produce a different (and not more understandable) error message afterwards

jfyuen:
> > > I wonder if we need to initialize `_internal_dict` if `data is None` as the class will be totally unusable in that case.
> > 
> > 
> > I wondered the same thing. I think it all depends if we want to raise an error on instance creation or when accessing the `dict`. I would prefer to test for `None` at dict creation, but that doesn't prevent using an empty `dict` that would fail later. There is no good way to check everything at object creation...
> 
> I agree with you, there is perfect check. But here, creating an empty dict does not bring anything IMO compared to just initialize the _internal_dict with `None`, it will just produce a different (and not more understandable) error message afterwards

Yes, we're on the same line. Done.

PR #68, opened 2020-04-14
Add methods to delete infra / delete a node on static infra

AgatheG:
[ch37697](https://app.clubhouse.io/dataiku/story/37697/improve-python-public-api-client-for-api-deployer)

https://github.com/dataiku/dip/pull/10420 should be merged first

### DIFF
--- a/dataikuapi/dss/apideployer.py
+++ b/dataikuapi/dss/apideployer.py
@@ -171,7 +171,7 @@ class DSSAPIDeployerInfra(object):
         Deletes this infra
         You may only delete an infra if it has no deployments on it anymore.
         """
-        return self.client._perform_empty(
+        self.client._perform_empty(
             "DELETE", "/api-deployer/infras/%s" % (self.infra_id))
 
 
@@ -207,10 +207,10 @@ class DSSAPIDeployerInfraSettings(object):
 
         :param str node_url: URL of the node to remove
         """
-        for node in self.settings["apiNodes"]:
-            if node["url"] == node_url:
+        api_nodes = list(self.settings["apiNodes"])
+        for node in api_nodes:
+            if node.get("url") == node_url:
                 self.settings["apiNodes"].remove(node)
-                break
 
     def get_raw(self):
         """
--- a/dataikuapi/dss/apideployer.py
+++ b/dataikuapi/dss/apideployer.py
@@ -166,6 +166,15 @@ class DSSAPIDeployerInfra(object):
 
         return DSSAPIDeployerInfraSettings(self.client, self.infra_id, settings)
 
+    def delete(self):
+        """
+        Deletes this infra
+        You may only delete an infra if it has no deployments on it anymore.
+        """
+        return self.client._perform_empty(
+            "DELETE", "/api-deployer/infras/%s" % (self.infra_id))
+
+
 class DSSAPIDeployerInfraSettings(object):
     """The settings of an API Deployer Infra. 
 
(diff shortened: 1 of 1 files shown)


dataikuapi/dss/apideployer.py
```diff
@@ -166,6 +166,15 @@ def get_settings(self):
 
         return DSSAPIDeployerInfraSettings(self.client, self.infra_id, settings)
 
+    def delete(self):
+        """
+        Deletes this infra
+        You may only delete an infra if it has no deployments on it anymore.
+        """
+        return self.client._perform_empty(
```
spicquenot, line 174: Does _perform_empty return anything?

dataikuapi/dss/apideployer.py
```diff
@@ -191,6 +200,18 @@ def add_apinode(self, url, api_key, graphite_prefix=None):
...
 
+    def remove_apinode(self, node_url):
+        """
+        Removes a node from the list of nodes of this infra.
+        Only applicable to STATIC infrastructures
+
+        :param str node_url: URL of the node to remove
+        """
+        for node in self.settings["apiNodes"]:
+            if node["url"] == node_url:
```
spicquenot, line 211: URL-empty proof by using .get("url", None), prevent some errors.

dataikuapi/dss/apideployer.py
```diff
@@ -191,6 +200,18 @@ def add_apinode(self, url, api_key, graphite_prefix=None):
...
+                self.settings["apiNodes"].remove(node)
+                break
```
spicquenot, line 213: To discuss: maybe do not break here to remove all the nodes which url matches. There is no unique check on URL of apiNodes

dataikuapi/dss/apideployer.py
instanceofme: To be fair, we have a mix of return / not return in the API client on `_perform_empty`

AgatheG: Nope :) I guess we can drop the return. I did it this way as it is done like pretty much everywhere else for delete endpoints but it is not a very good reason 😅

dataikuapi/dss/apideployer.py
AgatheG: It was my first move, but in `add_apinode` brackets are used instead of `get` as well. Actually we could even try/except the KeyError and have a message to explain that the infra is of the wrong type?

dataikuapi/dss/apideployer.py
AgatheG: Oops, I was persuaded URL was unique 🙂

dataikuapi/dss/apideployer.py
instanceofme: Meh, no need to go that far, would introduce tight coupling. Using `get` is error free and will do what is requested (if the infra has the wrong type, you still removed every node with that url, even if it means none).

AgatheG: Actually I read to fast and thought we were speaking of `self.settings["apiNodes"]`, do we want to use `get` here as well?

spicquenot: apiNodes is defaulted to be empty list for a static infra settings, so no need to change that, the content of the list, at contrary, can be anything since it's a user input

dataikuapi/dss/apideployer.py
spicquenot: Oh yes, I forgot python can "return" an "empty" value, nvm then

AgatheG: Too late the fix is pushed 😛

dataikuapi/dss/apideployer.py
AgatheG: Only if the infra is a static one and not a K8S though right? But well the user is not supposed to use this if his infra is not staticanyway so this might be overkill

instanceofme: That's in line with what's done in add, so indeed let's not overthink it

PR #76, opened 2020-05-25
Adding delete_versions to the saved model API.

lpenet:
Adding a `delete_versions` method to `DSSSavedModel`.

See https://github.com/dataiku/dip/pull/10513 for background.

### DIFF
--- a/dataikuapi/dss/savedmodel.py
+++ b/dataikuapi/dss/savedmodel.py
@@ -1,10 +1,8 @@
-from ..utils import DataikuException
-from ..utils import DataikuUTF8CSVReader
-from ..utils import DataikuStreamedHttpUTF8CSVReader
-import json
 from .ml import DSSTrainedPredictionModelDetails, DSSTrainedClusteringModelDetails
 from .metrics import ComputedMetrics
-from .discussion import DSSObjectDiscussions
+from .ml import DSSTrainedClusteringModelDetails
+from .ml import DSSTrainedPredictionModelDetails
+
 
 class DSSSavedModel(object):
     """
@@ -79,12 +77,13 @@ class DSSSavedModel(object):
         model, an intermediate version is created every time a partition has finished training.
         :type remove_intermediate: bool
         """
+        body = {
+            "versions": versions,
+            "removeIntermediate": remove_intermediate
+        }
         self.client._perform_empty(
             "POST", "/projects/%s/savedmodels/%s/actions/delete-versions" % (self.project_key, self.sm_id),
-            params={
-                'versions': versions,
-                'removeIntermediate': remove_intermediate
-            })
+            body=body)
     ########################################################
     # Metrics
     ########################################################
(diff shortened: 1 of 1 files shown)


dataikuapi/dss/savedmodel.py
```diff
@@ -69,6 +69,22 @@ def set_active_version(self, version_id):
...
+        :param versions: list of versions to delete
+        :type versions: list[str]
+        :param remove_intermediate: also remove intermediate versions (default: True). In the case of a partitioned
+        model, an intermediate version is created every time a partition has finished training.
+        :type remove_intermediate: bool
+        """
+        self.client._perform_empty(
+            "POST", "/projects/%s/savedmodels/%s/actions/delete-versions" % (self.project_key, self.sm_id),
+            params={
+                'versions': json.dumps(versions),
```
instanceofme, line 85: Code smell: JSON in JSON. See remarks on main PR

dataikuapi/dss/savedmodel.py
```diff
@@ -69,6 +69,22 @@ def set_active_version(self, version_id):
...
+        :param versions: list of versions to delete
+        :type versions: list[str]
+        :param remove_intermediate: also remove intermediate versions (default: True). In the case of a partitioned
+        model, an intermediate version is created every time a partition has finished training.
+        :type remove_intermediate: bool
+        """
+        self.client._perform_empty(
+            "POST", "/projects/%s/savedmodels/%s/actions/delete-versions" % (self.project_key, self.sm_id),
+            params={
+                'versions': versions,
```
instanceofme, line 85: In those helpers, we tend to be liberal in what we accept, so we can do a special treatment if `versions` is a string\* and automatically wrap it in an array.
\* Simpler to test as "not an array" because string is complicated (`unicode` or `str` in py2, `str` or `bytes` in py3…).

lpenet: I just checked: the JSON parser is smart enough to convert a string param to a list of strings with a single instance.
So, no need to complexify the client, IMHO.

instanceofme: Oh, nice

dataikuapi/dss/savedmodel.py
```diff
@@ -69,6 +67,23 @@ def set_active_version(self, version_id):
...
+        Delete version(s) of the saved model
+
+        :param versions: list of versions to delete
+        :type versions: list[str]
+        :param remove_intermediate: also remove intermediate versions (default: True). In the case of a partitioned
+        model, an intermediate version is created every time a partition has finished training.
+        :type remove_intermediate: bool
+        """
+        body = {
+            "versions": versions,
```
instanceofme, line 81: Initial remark stands then: should auto-wrap versions if isn't an array.

lpenet: I do not think so. It is dealt with server-side, by the custom deserializer.

lpenet: ok, so, following our discussion:
- the server side deserializer is dumber (but less complex)
- the ability to use a string instead of an array for a single version is handled in the wrapper, but not advertised.

PR #79, opened 2020-06-09
Added redetect flag to ML task guess public API

andreybavt:
Required for the DKU App to be able to use python API to remap ML task's target without changing the rest of the settings

Linked to https://github.com/dataiku/dip/pull/10681

### DIFF
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -1690,6 +1690,7 @@ class DSSMLTask(object):
         """
         Guess the feature handling and the algorithms.
         :param string prediction_type: In case of a prediction problem the prediction type can be specify. Valid values are BINARY_CLASSIFICATION, REGRESSION, MULTICLASS.
+        :param bool redetect: if True then the DSS automatically initializes ML task settings, if False, only the target values are remapped
         """
         obj = {}
         if prediction_type is not None:
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -1686,7 +1686,7 @@ class DSSMLTask(object):
             "POST", "/projects/%s/models/lab/%s/%s/models/%s/actions/redeployToFlow" % (self.project_key, self.analysis_id, self.mltask_id, model_id),
             body = obj)
 
-    def guess(self, prediction_type=None):
+    def guess(self, prediction_type=None, redetect=None):
         """
         Guess the feature handling and the algorithms.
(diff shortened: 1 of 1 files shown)


dataikuapi/dss/ml.py
```diff
@@ -1686,15 +1686,17 @@ def redeploy_to_flow(self, model_id, recipe_name=None, saved_model_id=None, acti
             "POST", "/projects/%s/models/lab/%s/%s/models/%s/actions/redeployToFlow" % (self.project_key, self.analysis_id, self.mltask_id, model_id),
             body = obj)
 
-    def guess(self, prediction_type=None):
+    def guess(self, prediction_type=None, redetect=None):
         """
         Guess the feature handling and the algorithms.
         :param string prediction_type: In case of a prediction problem the prediction type can be specify. Valid values are BINARY_CLASSIFICATION, REGRESSION, MULTICLASS.
+        :param bool redetect: if True then the DSS automatically initializes ML task settings, if False, only the target values are remapped
```
instanceofme, line 1693: Should specify that `True` is the default. Can also skip the `True` part, kind of obvious for a boolean.

```suggestion
        :param bool redetect: if False then skip ML task settings automatic detection, only the target values are remapped
```

dataikuapi/dss/ml.py
```diff
@@ -1686,18 +1686,18 @@ def redeploy_to_flow(self, model_id, recipe_name=None, saved_model_id=None, acti
             "POST", "/projects/%s/models/lab/%s/%s/models/%s/actions/redeployToFlow" % (self.project_key, self.analysis_id, self.mltask_id, model_id),
             body = obj)
 
-    def guess(self, prediction_type=None, target_only=None):
+    def guess(self, prediction_type=None, reguess_level=None):
         """
         Guess the feature handling and the algorithms.
         :param string prediction_type: In case of a prediction problem the prediction type can be specify. Valid values are BINARY_CLASSIFICATION, REGRESSION, MULTICLASS.
-        :param bool target_only: if True then skip ML task settings automatic detection, only the target values are remapped. Only valid for prediction ML Tasks.
+        :param bool reguess_level: One of the following values: TARGET_CHANGE, TARGET_REGUESS and FULL_REGUESS. Only valid for prediction ML Tasks.
```
instanceofme, line 1693:
```suggestion
        :param bool reguess_level: One of the following values: TARGET_CHANGE, TARGET_REGUESS and FULL_REGUESS. Only valid for prediction ML Tasks, cannot be specified if prediction_type is also set.
```

PR #80, opened 2020-06-10
Enhancement/dss80 write file mf subfolder

chaydont:
[ch50316](https://app.clubhouse.io/dataiku/story/50316)

### DIFF
--- a/dataikuapi/dss/utils.py
+++ b/dataikuapi/dss/utils.py
@@ -47,7 +47,13 @@ class DSSComputedColumn(object):
     def formula(name, formula, type="double"):
         return {"expr": formula, "mode": "GREL", "name": name, "type": type}
 
-from enum import Enum
+import sys
+if sys.version_info > (3,4):
+    from enum import Enum
+else:
+    class Enum(object):
+        pass
+
 class DSSFilterOperator(Enum):
     EMPTY_ARRAY = "empty array"
     NOT_EMPTY_ARRAY = "not empty array"


dataikuapi/dss/managedfolder.py
```diff
@@ -90,18 +90,18 @@ def delete_file(self, path):
...
         Args:
             f: the file contents, as a stream
-            name: the name of the file
+            path: the path of the file
         """
 
         return self.client._perform_json_upload(
-                "POST", "/projects/%s/managedfolders/%s/contents/" % (self.project_key, self.odb_id),
-                name, f)
+                "POST", "/projects/%s/managedfolders/%s/contents/%s" % (self.project_key, self.odb_id, path),
```
fterrazzoni, line 103: 1. Appending raw path to the URL won't always work because of URL encoding/decoding rules. For instance, if the filename contains a `%` character it will either fail, or interpret it like an encoded character (eg. `%20` will become a space, etc)
2. This function raises `UnicodeEncodeError: 'ascii' codec can't encode characters` if non-ascii characters are present in the path


You may want to use a function like `requests.utils.quote(...)` to fix (1). Luckily, it will also solve (2) at the same time because special characters will be replaced by percent-encoded entities

tests/managedfolder_tests.py
```diff
@@ -57,12 +57,12 @@ def updload_download_test():
 
     count = len(managedfolder.list_contents()['items'])
     with open(stuff, "r") as f:
-        managedfolder.put_file('stuff', f)
+        managedfolder.put_file('subfolder/stuff', f)
```
fterrazzoni, line 60: These test files are un-used and have not been updated since 5+ years :-) 

In order to test your modification, you need to add a test in the main DSS repository (along with other integration tests). This is how we currently these the Python client

Note: we should open a PR to remove these test files because you're not the first person to get confused by them

fterrazzoni: Also, this test could detect more errors if it was adding special characters (accents, percents, ...), since it is a frequent problem with Python 2/3

fterrazzoni: FYI I remove these tests in https://github.com/dataiku/dataiku-api-client-python/commit/f9fb9881fd1db591bfc29a2e41ca65dc941318a5 so that people don't try to update them in the future

PR #82, opened 2020-06-19
Add two methods to facilitate navigation from SM's objects to analysis' objects

LouisPlt:
[ch47819](https://app.clubhouse.io/dataiku/story/47819/api-to-get-source-analysis-from-saved-model-version) The goal here is to mimic the two buttons in the UI "View original analysis". 
The first on is located in the saved model page (where the list of versions are) and it redirects to the last exported ML task.
The second is located in the saved model version page and it redirects to the origin trained model of this version (either if it has been export or either if it has been retraining )

 Add two methods to facilitate navigation from saved model's objects to analysis' objects:
- for getting the origin mltask of a saved model:
```python
mltask:DSSMLTask = saved_model.get_origin_ml_task()
```
- for getting the origin trained model of a saved model version:
```python
trained_model:DSSTrainedModelDetails = saved_model_version.get_origin_analysis_trained_model()
```

### DIFF
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -1,4 +1,5 @@
-from dataikuapi.dss.utils import extract_info_from_full_model_id
+import re
+
 from ..utils import DataikuException
 from ..utils import DataikuUTF8CSVReader
 from ..utils import DataikuStreamedHttpUTF8CSVReader
@@ -537,9 +538,8 @@ class DSSTrainedModelDetails(object):
             if "smOrigin" not in self.get_raw():
                 raise DataikuException("Unknow ")
             fmi = self.get_raw()["smOrigin"]["fullModelId"]
-            _, analysis_id, mltask_id = extract_info_from_full_model_id(fmi)
-
-            origin_ml_task = DSSMLTask(self.saved_model.client, self.saved_model.project_key, analysis_id, mltask_id)
+            origin_ml_task = DSSMLTask.from_full_model_id(self.saved_model.client, fmi,
+                                                          project_key=self.saved_model.project_key)
             return origin_ml_task.get_trained_model_details(fmi)
 
 class DSSTreeNode(object):
@@ -1443,6 +1443,14 @@ class DSSTrainedClusteringModelDetails(DSSTrainedModelDetails):
 
 
 class DSSMLTask(object):
+
+    @staticmethod
+    def from_full_model_id(client, fmi, project_key=None):
+        match = re.match(u'A-([\w]+)-([\w]+)-([\w]+)-s\w+-pp\w+-m\w+', fmi)
+        if project_key is None:
+            project_key = match.group(1)
+        return DSSMLTask(client, project_key, match.group(2), match.group(3))
+
     """A handle to interact with a MLTask for prediction or clustering in a DSS visual analysis"""
     def __init__(self, client, project_key, analysis_id, mltask_id):
         self.client = client
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -528,19 +528,19 @@ class DSSTrainedModelDetails(object):
 
     def get_origin_analysis_trained_model(self):
         """
-        Fetch details about the model in an analysis, this model has been exported from
+        Fetch details about the model in an analysis, this model has been exported from. Returns None if the
+        deployed trained model does not have a origin analysis trained model.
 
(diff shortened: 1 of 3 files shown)


dataikuapi/dss/utils.py
```diff
@@ -257,4 +260,9 @@ def custom_fields(self):
...
     @custom_fields.setter
     def custom_fields(self, custom_fields):
-        self._tod["customFields"] = custom_fields
\ No newline at end of file
+        self._tod["customFields"] = custom_fields
+
+
+def extract_info_from_full_model_id(full_model_id):
+    match = re.search(u'([\w]+)-([\w]+)-([\w]+)-s\w+-pp\w+-m\w+', full_model_id)
+    return match.group(1), match.group(2), match.group(3)
```
instanceofme, line 268: - Only used once, can inline for now. A proper generic function would need to be more though out, potentially use capturing group names and return the full structure, or even a custom object equivalent to backend `FullModelId`, in which case this would be a static factory method…
- In context, this can only return analysis trained models, so it should
    - match the whole string
    - enforce the initial `A-`

LouisPlt: I'm not sure I understand all, but this method is used twice.

instanceofme: Ah true, but in both cases we treat it exactly the same. In this case, you can move this as `MLTask.from_full_model_id(fmi)`. (And second point still stands)

instanceofme: If we want to override project key, can add that as an optional argument

dataikuapi/dss/ml.py
```diff
@@ -1443,6 +1443,14 @@ def get_scatter_plots(self):
 
 
 class DSSMLTask(object):
+
+    @staticmethod
+    def from_full_model_id(client, fmi, project_key=None):
+        match = re.match(u'A-([\w]+)-([\w]+)-([\w]+)-s\w+-pp\w+-m\w+', fmi)
```
instanceofme, line 1449: Doesn't work with partitioned models, see `FullModelId.ANALYSIS_PATTERN`
```suggestion
        match = re.match(u'A-([\w]+)-([\w]+)-([\w]+)-s\d+-pp\d+(?:-base|-part-\w+)?-m\d+', fmi)
```
(suggestion needs to be tested)

dataikuapi/dss/savedmodel.py
```diff
@@ -100,8 +99,9 @@ def get_origin_ml_task(self):
         """
         definition = self.get_definition()
         fmi = definition["lastExportedFrom"]
-        _, analysis_id, ml_task_id = extract_info_from_full_model_id(fmi)
-        return DSSMLTask(self.client, self.project_key, analysis_id, ml_task_id)
+
+        origin_ml_task = DSSMLTask.from_full_model_id(self.client, fmi, project_key=self.project_key)
+        return origin_ml_task.get_trained_model_details(fmi)
```
instanceofme: I guess we could manage the case where it's not present by returning None? e.g.

```python
fmi = self.get_definition().get("lastExportedFrom", None)
if fmi is not None:
    return DSSMLTask.from_full_model_id(self.client, fmi, project_key=self.project_key).get_trained_model_details(fmi)
```

dataikuapi/dss/ml.py
```diff
@@ -537,9 +538,8 @@ def get_origin_analysis_trained_model(self):
             if "smOrigin" not in self.get_raw():
                 raise DataikuException("Unknow ")
             fmi = self.get_raw()["smOrigin"]["fullModelId"]
-            _, analysis_id, mltask_id = extract_info_from_full_model_id(fmi)
-
-            origin_ml_task = DSSMLTask(self.saved_model.client, self.saved_model.project_key, analysis_id, mltask_id)
+            origin_ml_task = DSSMLTask.from_full_model_id(self.saved_model.client, fmi,
+                                                          project_key=self.saved_model.project_key)
```
instanceofme: Ditto

dataikuapi/dss/savedmodel.py
```diff
@@ -1,3 +1,4 @@
+from dataikuapi.dss.ml import DSSMLTask
 from .ml import DSSTrainedPredictionModelDetails, DSSTrainedClusteringModelDetails
 from .metrics import ComputedMetrics
```
instanceofme, line 3:
```suggestion
from .metrics import ComputedMetrics
from .ml import DSSMLTask
```

PR #83, opened 2020-06-25
provide path to put_file to write in subfolder

chaydont:
Rebirth of #80 
[ch50316](https://app.clubhouse.io/dataiku/story/50316)

### DIFF
--- a/dataikuapi/dss/managedfolder.py
+++ b/dataikuapi/dss/managedfolder.py
@@ -2,6 +2,7 @@ from ..utils import DataikuException
 from ..utils import DataikuUTF8CSVReader
 from ..utils import DataikuStreamedHttpUTF8CSVReader
 import json
+from requests import utils
 from .metrics import ComputedMetrics
 from .future import DSSFuture
 from .discussion import DSSObjectDiscussions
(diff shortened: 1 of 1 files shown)


dataikuapi/dss/managedfolder.py
```diff
@@ -81,27 +82,26 @@ def get_file(self, path):
...
         Args:
             f: the file contents, as a stream
-            name: the name of the file
+            path: the path of the file
         """
-
         return self.client._perform_json_upload(
-                "POST", "/projects/%s/managedfolders/%s/contents/" % (self.project_key, self.odb_id),
-                name, f)
+                "POST", "/projects/%s/managedfolders/%s/contents/%s" % (self.project_key, self.odb_id, utils.quote(path)),
```
fterrazzoni, line 103: I've noticed  `path` is passed twice:
1. First one goes to URI and is used by DSS (primarily used by DSS)
2. Second one is used for the `Content-Disposition:` header of the attached file (used as fallback by DSS when former is empty)

The sad part now is that while (1) correctly handles special character after you fix, (2) breaks with special characters (such as accents). I suggest:
- Adding a check ensuring `path` is non-empty (to ensure we're always in situation (1))
- Passing a static/dummy path (eg;  `tmpdata`) - it will be ignored by DSS

chaydont: If we do that there could still be situations where it creates a file `tmpdata` except of raising an exception and we probably dont want that. For example when calling `mf.put_file("/", f)`.

Providing the path also in the header shouldn't break any special characters because if there are special characters, the path isn't empty, so it will take the path from (1).

If you know a simple way to ensure than the path given will be detected as a path by `extractPathFromRequest` in the backend, it seem cleaner to do it your way, but if not, I think keeping it this way would work as expected in any situation, wdyt?

chaydont: or maybe we can just keep the path in the header empty, would probably make more sense

fterrazzoni: I've not tried and I don't know if it's valid

PR #84, opened 2020-06-29
Fix/ml weighting

samronsin:
Sorry for polluting history on master...

Found and fixed 2 bugs (cc @kevanescence regarding [CH51777](https://app.clubhouse.io/dataiku/story/51777/investigate-drop-rows-tests-on-weight-test-failure)):
- The feature selected to be sample weight in `set_weighting` is not the proper one because of a local variable name conflict with `feature_name`
- Using `remove_sample_weighting` raises an exception as it calls an unimplemented method `unset_weighting`

### DIFF
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -460,10 +460,10 @@ class DSSPredictionMLTaskSettings(DSSMLTaskSettings):
 
     def remove_sample_weighting(self):
         """
-        Deprecated. Use unset_weighting() instead
+        Deprecated. Use set_weighting(method=\"NO_WEIGHTING\") instead
         """
-        warnings.warn("remove_sample_weighting() is deprecated, please use set_weigthing(method=\"NO_WEIGHTING\") instead", DeprecationWarning)
-        return self.set_weigthing(method="NO_WEIGHTING")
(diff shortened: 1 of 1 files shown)


dataikuapi/dss/ml.py
```diff
@@ -463,7 +463,8 @@ def remove_sample_weighting(self):
         Deprecated. Use unset_weighting() instead
         """
         warnings.warn("remove_sample_weighting() is deprecated, please use set_weigthing(method=\"NO_WEIGHTING\") instead", DeprecationWarning)
-        return self.unset_weighting()
+        return self.set_weigthing(method="NO_WEIGHTING")
```
nicolasservel, line 466:
```suggestion
        return self.set_weighting(method="NO_WEIGHTING")
```

dataikuapi/dss/ml.py
```diff
@@ -463,7 +463,8 @@ def remove_sample_weighting(self):
         Deprecated. Use unset_weighting() instead
         """
         warnings.warn("remove_sample_weighting() is deprecated, please use set_weigthing(method=\"NO_WEIGHTING\") instead", DeprecationWarning)
```
nicolasservel, line 465:
```suggestion
        warnings.warn("remove_sample_weighting() is deprecated, please use set_weighting(method=\"NO_WEIGHTING\") instead", DeprecationWarning)
```

nicolasservel:
Besides, this PR should probably target `release/8.0`

PR #86, opened 2020-06-30
Add DSSSavedModelSettings class

LouisPlt:
[ch47819](https://app.clubhouse.io/dataiku/story/47819/api-to-get-source-analysis-from-saved-model-version): Use the modern API pattern for `DSSSavedModel` by replacing the `get_definition`  by a new class: `DSSSavedModelSettings` and its method `get_raw`

### DIFF
--- a/dataikuapi/dss/savedmodel.py
+++ b/dataikuapi/dss/savedmodel.py
@@ -20,9 +20,16 @@ class DSSSavedModel(object):
     def id(self):
         return self.sm_id
 
-    def get_definition(self):
-        return self.client._perform_json(
+    def get_settings(self):
+        """
+        Returns the settings of this saved model.
+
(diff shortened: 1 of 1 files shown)


dataikuapi/dss/savedmodel.py
```diff
@@ -173,3 +179,13 @@ def delete(self):
 
         """
         return self.client._perform_empty("DELETE", "/projects/%s/savedmodels/%s" % (self.project_key, self.sm_id))
+
+
+class DSSSavedModelSettings:
+
```
instanceofme, line 185:
```suggestion
    """
    A handle on the settings of a saved model

    Do not create this class directly, instead use :meth:`dataikuapi.dss.DSSSavedModel.get_settings`
    """

```

PR #88, opened 2020-07-07
Python2/3 compatilibity

adescamps:
[[ch52560]](https://app.clubhouse.io/dataiku/story/52560)


adescamps:
Not sure between that and `isinstance(zone, str) or isinstance(zone, unicode)` which is better?

cstenac:
"unicode" does not exist in Python 3 either, so that would not work.

Did not test but code LGTM

PR #96, opened 2020-09-21
Add Python client for Project deployer

AgatheG:
[ch47130](https://app.clubhouse.io/dataiku/story/47130/public-api-for-project-deployer)

See PR for the Public API on dip: https://github.com/dataiku/dip/pull/11101

### DIFF
--- a/dataikuapi/dss/apideployer.py
+++ b/dataikuapi/dss/apideployer.py
@@ -1,6 +1,7 @@
 import json
 from .future import DSSFuture
 
+
 class DSSAPIDeployer(object):
     """
     Handle to interact with the API Deployer.
@@ -10,7 +11,7 @@ class DSSAPIDeployer(object):
     def __init__(self, client):
         self.client = client
 
-    def list_deployments(self, as_objects = True):
+    def list_deployments(self, as_objects=True):
         """
         Lists deployments on the API Deployer
 
@@ -97,7 +98,7 @@ class DSSAPIDeployer(object):
         """
         return DSSAPIDeployerInfra(self.client, infra_id)
 
-    def list_services(self, as_objects = True):
+    def list_services(self, as_objects=True):
         """
         Lists API services on the API Deployer
 
--- a/dataikuapi/dss/apideployer.py
+++ b/dataikuapi/dss/apideployer.py
@@ -56,7 +56,7 @@ class DSSAPIDeployer(object):
         self.client._perform_json("POST", "/api-deployer/deployments", body=settings)
         return self.get_deployment(deployment_id)
 
-    def list_infras(self, as_objects = True):
+    def list_infras(self, as_objects=True):
         """
         Lists deployment infrastructures on the API Deployer
 
--- a/dataikuapi/dss/apideployer.py
+++ b/dataikuapi/dss/apideployer.py
@@ -56,6 +56,15 @@ class DSSAPIDeployer(object):
         self.client._perform_json("POST", "/api-deployer/deployments", body=settings)
         return self.get_deployment(deployment_id)
 
+    def list_stages(self):
+        """
+        Lists infrastructure stages of the API Deployer
+
+        :rtype: a list of dict. Each dict contains a field "id" for the stage identifier and "desc" for its description.
+        :rtype: list
+        """
+        return self.client._perform_json("GET", "/api-deployer/stages")
+
     def list_infras(self, as_objects=True):
         """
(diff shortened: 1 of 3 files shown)


dataikuapi/dss/projectdeployer.py
```diff
@@ -0,0 +1,450 @@
+from .future import DSSFuture
+
+class DSSProjectDeployer(object):
```
mhoarau, line 3: Just a few code style issues: here we miss a blank line, later we have spaces around a `=` for default values, ...

dataikuapi/dss/projectdeployer.py
```diff
@@ -0,0 +1,450 @@
...
+        if origin_project_key is None and design_node_url is None and design_node_id is None:
+            params = None
+        else:
+            params = {
+                "projectKey": origin_project_key,
+                "nodeId": design_node_id,
+                "url": design_node_url
+            }
+        return self.client._perform_empty("POST",
+                "/project-deployer/projects/%s/packages" % (self.project_key), params=params, files={"file":fp})
```
mhoarau, line 380: The correct endpoint is `bundles`, not `packages`.

dataikuapi/dss/projectdeployer.py
```diff
@@ -0,0 +1,450 @@
+from .future import DSSFuture
```
mhoarau, line 1: Is it on purpose that we don't expose:
* `com.dataiku.dip.server.api.PublicAPIProjectDeployerController#deleteProject`
* `com.dataiku.dip.server.api.PublicAPIProjectDeployerController#listStages` 

?

mhoarau (changes requested):

dataikuapi/dss/projectdeployer.py
AgatheG: For now I merely followed the methods we had in `apideployer.py`, "translating" them to Project deployer methods. For `deleteProject` especially, there was no endpoint in the Public API to delete services before dataiku/dip#11101. We can consider adding those to both deployers? 🙂

dataikuapi/dss/projectdeployer.py
AgatheG: Should be OK after ab1dfc1, if I missed some please say so 😁

instanceofme (approved):

dataikuapi/dss/projectdeployer.py
```diff
@@ -0,0 +1,451 @@
...
+        if origin_project_key is None and design_node_url is None and design_node_id is None:
+            params = None
+        else:
+            params = {
+                "projectKey": origin_project_key,
+                "nodeId": design_node_id,
+                "url": design_node_url
+            }
+        return self.client._perform_empty("POST",
+                "/project-deployer/projects/%s/bundles" % (self.project_key), params=params, files={"file":fp})
```
marchelleboid, line 381: If you try to send any of the optional params, they won't be received on the DSS side: `designNodeInfo` will be `null` as the controller is looking for a request param with that name. One way to fix this is by doing `params={"designNodeInfo": json.dumps(params)}` (and `import json` at the top), which leads to a pretty ugly request. Maybe it's better to change the controller method to accept each of those three optional params individually? In that case, this code could remain unchanged, you would just need to build the `DesignNodeInfo` object manually in the controller method in the public API PR.

marchelleboid (changes requested):

dataikuapi/dss/apideployer.py
```diff
@@ -10,7 +11,7 @@ class DSSAPIDeployer(object):
     def __init__(self, client):
         self.client = client
 
-    def list_deployments(self, as_objects = True):
+    def list_deployments(self, as_objects=True):
```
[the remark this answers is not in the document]
AgatheG: Nice catch!

dataikuapi/dss/projectdeployer.py
AgatheG: Done in c1ed433, corresponding changes on the Public API endpoint in a86ba14

dataikuapi/dss/projectdeployer.py
AgatheG: Done in 6a220bd

marchelleboid (approved):
LGTM (only comment is the naming of the params on the other PR)

mhoarau (approved):
LGTM (I did not extensively played with all the APIs though)

PR #98, opened 2020-10-06
Add functions to export a wiki article as a PDF

chaydont:
[ch55059](https://app.clubhouse.io/dataiku/story/55059)
Need an instance with at least [dip#11418](https://github.com/dataiku/dip/pull/11418) merged (not on master yet)

### DIFF
--- a/dataikuapi/dss/wiki.py
+++ b/dataikuapi/dss/wiki.py
@@ -89,6 +89,30 @@ class DSSWiki(object):
 
         return article
 
+    def get_export_stream(self, paperSize="A4", exportAttachment=False):
+        """
+        Download the whole wiki of the project in PDF format as a binary stream.
+        Warning: this stream will monopolize the DSSClient until closed.
+        """
+        body = {
+            "paperSize": paperSize,
+            "exportAttachment": exportAttachment
+        }
+        return self.client._perform_raw("POST", "/projects/%s/wiki/actions/export" % (self.project_key, self.article_id), body=body)
+
+    def export_to_file(self, path, paperSize="A4", exportAttachment=False):
+        """
+        Download the whole wiki of the project in PDF format into the given output file.
+        """
+        stream = self.get_export_stream(paperSize=paperSize, exportAttachment=exportAttachment)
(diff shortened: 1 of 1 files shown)


dataikuapi/dss/wiki.py
```diff
@@ -89,6 +89,30 @@ def create_article(self, article_name, parent_id=None, content=None):
...
+            "paperSize": paperSize,
+            "exportAttachment": exportAttachment
+        }
+        return self.client._perform_raw("POST", "/projects/%s/wiki/actions/export" % (self.project_key), body=body)
+
+    def export_to_file(self, path, paperSize="A4", exportAttachment=False):
+        """
+        Download the whole wiki of the project in PDF format into the given output file.
+        """
+        stream = self.get_export_stream(paperSize=paperSize, exportAttachment=exportAttachment)
```
FChataigner, line 107: is it possible to use `with ... as stream:` instead of the get/do/close construction? (if the stream can handle it)

dataikuapi/dss/wiki.py
```diff
@@ -247,6 +271,32 @@ def get_uploaded_file(self, upload_id):
...
+            "exportChildren": exportChildren,
+            "exportAttachment": exportAttachment
+        }
+        return self.client._perform_raw("POST", "/projects/%s/wiki/%s/actions/export" % (self.project_key, self.article_id), body=body)
+
+    def export_to_file(self, path, paperSize="A4", exportChildren=False, exportAttachment=False):
+        """
+        Download an article in PDF format into the given output file.
+        """
+        stream = self.get_export_stream(paperSize=paperSize, exportChildren=exportChildren, exportAttachment=exportAttachment)
```
FChataigner, line 290: ditto

FChataigner (changes requested):
looks good

dataikuapi/dss/wiki.py
chaydont: I added it, it's indeed neater. Also I used underscores instead of camelCase for parameter names

PR #100, opened 2020-10-14
diagnostics: add helper in DSSTrainedModelDetails

jfyuen:
Add a helper to easily access and use Hints

### DIFF
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -543,6 +543,62 @@ class DSSTrainedModelDetails(object):
                                                               project_key=self.saved_model.project_key)
                 return origin_ml_task.get_trained_model_details(fmi)
 
+    def get_hints(self):
+        """
+        Retrieve hints computed for this trained model
+
+        :returns: list of hints
+        :rtype: list of type `dataikuapi.dss.ml.DSSMLHint`
+        """
+        hints = self.details.get("trainHints", {})
+        return [DSSMLHint(d) for d in hints.get("hints", [])]
+
+
+class DSSMLHint(object):
+    """
+    Object that represents a computed Hint on a trained models
+
+    Do not create this object directly, use :meth:`DSSTrainedModelDetails.get_hints()` instead
+    """
+
+    def __init__(self, data):
+        self._internal_dict = data
+
+    def get_raw(self):
+        """
+        Gets the raw dictionary of the hint
+
+        :rtype: dict
+        """
+        return self._internal_dict
+
+    def get_type(self):
+        """
+        Return the base Hint type
+        :rtype: str
+        """
+        return self._internal_dict["type"]
+
+    def get_type_pretty(self):
+        """
+        Return the Hint type as displayed in the UI
+        :rtype: str
+        """
+        return self._internal_dict["displayableType"]
+
+    def get_message(self):
+        """
+        Return the message as displayed in the UI
+        :rtype: str
+        """
+        return self._internal_dict["message"]
+
+    def __repr__(self):
+        return "{cls}(type={type}, message={msg})".format(cls=self.__class__.__name__,
+                                                          type=self._internal_dict["type"],
+                                                          msg=self._internal_dict["message"])
+
+
 class DSSTreeNode(object):
     def __init__(self, tree, i):
         self.tree = tree
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -545,7 +545,7 @@ class DSSTrainedModelDetails(object):
 
     def get_hints(self):
         """
-        Retrieve hints computed for this trained model
+        Retrieves hints computed for this trained model
 
         :returns: list of hints
         :rtype: list of type `dataikuapi.dss.ml.DSSMLHint`
@@ -556,7 +556,7 @@ class DSSTrainedModelDetails(object):
 
 class DSSMLHint(object):
     """
-    Object that represents a computed Hint on a trained models
+    Object that represents a computed Hint on a trained model
 
     Do not create this object directly, use :meth:`DSSTrainedModelDetails.get_hints()` instead
     """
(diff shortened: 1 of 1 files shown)


nicolasservel (changes requested):
I think it would interesting to have a `DSSMLHints` that holds the list of triggered hints for the training:
* that can `group_by_type` to return a more user friendly object
* that can `get_hints_with_type`

Also, we need to enrich the documentation with the hint types so that the user can find his way around.

This PR should also add helpers to enable/disable hints from `module_utils.dataikuapi.dss.ml.DSSMLTaskSettings`

instanceofme:
> I think it would interesting to have a `DSSMLHints` that holds the list of triggered hints for the training:
> 
> * that can `group_by_type` to return a more user friendly object
> * that can `get_hints_with_type`

Not sure I agree, as this additional object is pure intermediate structure to get a grouped answer, I find it kind of gets in your way. Also, I didn't understand what `get_hints_with_type` would do?

IMO a list of hints is fine, but how about this: an option: `model.get_hints(group=None)`.
- `get_hints()` returns [hints]
- `get_hints(True)` return a dict {type → [hints]}
- in the future, so would `get_hints("type")` but we keep the possibility of later offering to group by other things, by not naming it `group_by_type`

nicolasservel:
`get_hints_with_type(HINT_TYPE)` -> returns list of hints of type `HINT_TYPE`

That can be optained, with your solution, by `get_hints(True)[HINT_TYPE]`

jfyuen:
> IMO a list of hints is fine, but how about this: an option: `model.get_hints(group=None)`.
> 
>     * `get_hints()` returns [hints]
> 
>     * `get_hints(True)` return a dict {type → [hints]}
> 
>     * in the future, so would `get_hints("type")` but we keep the possibility of later offering to group by other things, by not naming it `group_by_type`

I don't really like returing 2 different structures (`list` and `dict`) based on the parameter. Also grouping, like filtering, is really easy. Do we want to "complexify" the API for that?
If we want grouping, I would be for adding a dedicated method, like @nicolasservel suggested, so we would have like:
- `get_hints_by(group) -> dict`.
- `get_hints() -> list`

instanceofme:
We don't want to add too many methods either, as the API is already quite populated. IMO we could just have the one `get_hints()` for now

dataikuapi/dss/ml.py
```diff
@@ -253,6 +253,50 @@ def get_algorithm_settings(self, algorithm_name):
...
+        :param str hint_type: Name (in capitals) of the hint type.
+        :param bool enabled: if the hint should be enabled or not
+        """
+        settings = self.get_hints_settings()["settings"]
+        hint = [h for h in settings if h["type"] == hint_type]
+        if len(hint) == 0:
+            raise ValueError("hint type {} not found in settings".format(hint_type))
+        if len(hint) > 1:
+            raise ValueError("should not happen: multiple hint types {} not found in settings".format(hint_type))
+        hint[0]["enabled"] = enabled
```
instanceofme, line 298: - [Nitpick] capitalize
- multiple hints _not_ found?
- should probably simplify to something like

```python
hint = next(h for h in settings if h["type"] == hint_type, None)
if hint is None:
    raise ValueError("hint type {} not found in settings".format(hint_type))
hint["enabled"] = enabled
```

dataikuapi/dss/ml.py
```diff
@@ -253,6 +253,50 @@ def get_algorithm_settings(self, algorithm_name):
...
+
+        Please refer to the documentation for details on available hints.
+
+        :param str hint_type: Name (in capitals) of the hint type.
+        :param bool enabled: if the hint should be enabled or not
+        """
+        settings = self.get_hints_settings()["settings"]
+        hint = [h for h in settings if h["type"] == hint_type]
+        if len(hint) == 0:
+            raise ValueError("hint type {} not found in settings".format(hint_type))
```
instanceofme, line 295: What about old models? Are hint types consolidated by backend on the GET api call (or via default values)?

dataikuapi/dss/ml.py
```diff
@@ -253,6 +253,50 @@ def get_algorithm_settings(self, algorithm_name):
...
+        return self.mltask_settings["hintSettings"]
+
+    def set_hints_enabled(self, enabled):
+        """
+        Globally enables or disables all hints.
+
+        :param bool enabled: if the hints should be enabled or not
+        """
+        settings = self.get_hints_settings()
+        settings["enabled"] = enabled
```
instanceofme, line 281: [Nitpick] Can inline immediately-used-once
```suggestion
        self.get_hints_settings()["enabled"] = enabled
```

dataikuapi/dss/ml.py
```diff
@@ -253,6 +253,50 @@ def get_algorithm_settings(self, algorithm_name):
...
+    def get_hints_settings(self):
+        """
+        Gets the hints settings for a mltask. This returns a reference to the
+        hints' settings, not a copy, so changes made to the returned object will be reflected when saving.
+
+        This method returns a dictionary of the settings with:
+        - 'enabled': indicates if the hints are enabled globally, if False, all hints will be disabled
+        - 'settings': a dict comprised of:
+          - 'type': the hint type
+          - 'enabled': indicates if the hint type is enabled, if False, all hints of that type will be disabled
```
instanceofme, line 265: If it's a list of dict or a dict (what's the key then?)

instanceofme (changes requested):

dataikuapi/dss/ml.py
jfyuen: > multiple hints not found?

It is actually as you guessed `multiple hint types found`. I am being defensive, maybe too much, multiple hints of the same types here is clearly a bug and should not happen.

dataikuapi/dss/ml.py
jfyuen: Clarified: a list of dict

dataikuapi/dss/ml.py
jfyuen: Everything is on the backend with default values (all enabled), even for old models. Just tested with a project from 8.0.2 imported on an instance with hints.

dataikuapi/dss/ml.py
instanceofme: IMO the simplicity justifies being a bit less defensive

jfyuen: I think this method is already quite simple, we just filter on hint type and check the length. The defensiveness doesn't complexify the code that much.

nicolasservel (approved):
Should wait for renaming before merge, except that LGTM

instanceofme (approved):
No further remarks than what I already wrote

nicolasservel (approved):

PR #102, opened 2020-11-10
fix impersonation also in public API

thtrunck:
[ch46619](https://app.clubhouse.io/dataiku/story/46619/hiveexecutor-impalaexecutor-cannot-be-used-when-there-is-no-project-free-impersonation-rule)

This depends on: https://github.com/dataiku/dip/pull/11203

### DIFF
--- a/dataikuapi/dss/sqlquery.py
+++ b/dataikuapi/dss/sqlquery.py
@@ -7,7 +7,7 @@ class DSSSQLQuery(object):
     """
     A connection to a database or database-like on which queries can be run through DSS
     """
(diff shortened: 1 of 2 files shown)


FChataigner (changes requested):
you can't `import dataiku` in this package. Never, ever.

just put a fake project name like `__DKU_ANY_PROJECT__` as default

thtrunck:
> you can't `import dataiku` in this package. Never, ever.
> 
> just put a fake project name like `__DKU_ANY_PROJECT__` as default

I remove the default project.
Could you comment a bit more about why we shouldn't import dataiku here? Because `get_default_project` is doing exactly that, so technically I could extract the project_key from here. But I guess that's also not a good idea.

PR #103, opened 2020-11-13
Verify that the state does not already contain the result.

avigny:
Story details: [ch56069](https://app.clubhouse.io/dataiku/story/56069)

When the backend response is very fast to be computed, the initial response to the api call already contains the result, and does not contain a jobId.
This leads to the calls to get updates on the future to fail (no jobId)
The fix is to verify that the state does not already contain the result before an update.

### DIFF
--- a/dataikuapi/dss/future.py
+++ b/dataikuapi/dss/future.py
@@ -72,7 +72,7 @@ class DSSFuture(object):
         """
         Wait and get the future result
         """
-        if self.state.get('hasResult', False) and 'result' in self.state:
+        if self.state.get('hasResult', False):
             return self.result_wrapper(self.state.get('result', None))
         if self.state is None or not self.state.get('hasResult', False) or self.state_is_peek:
             self.get_state()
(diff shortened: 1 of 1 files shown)


dataikuapi/dss/future.py
```diff
@@ -72,6 +72,8 @@ def wait_for_result(self):
         """
         Wait and get the future result
         """
+        if self.state.get('hasResult', False) and 'result' in self.state:
```
fterrazzoni, line 75: I don't think we should have this additional `'result' in self.state` check:
- The source of truth telling whether or not the result is available is `hasResult`
- The presence of `result` is not guaranteed (even if the future is completed) because our JSON serializer drops null-valued properties (for instance `{"result": null}`  gets serialized as `{}` by DSS)

fterrazzoni (changes requested):

dataikuapi/dss/future.py
avigny: Yes, if `result not in self.state` then this is handled by `self.state.get('result', None)` and the default value will be `None`
👍

fterrazzoni (approved):

PR #106, opened 2020-11-20
Set output type to correct value when running training recipe

mhham:
# Why

The `DSSRecipe.run` method fails for (prediction and clustering) training recipes, because the recipe output type is not correctly set to `"SAVED_MODEL"`, hence interpreted as a dataset.

# What
If the recipe to run is a training recipe, correctly set the output type to `"SAVED_MODEL"`

[ch54893]

### DIFF
--- a/dataikuapi/dss/recipe.py
+++ b/dataikuapi/dss/recipe.py
@@ -76,18 +76,25 @@ class DSSRecipe(object):
         :return: the :class:`dataikuapi.dss.job.DSSJob` job handle corresponding to the built job
         :rtype: :class:`dataikuapi.dss.job.DSSJob`
         """
+        project = self.client.get_project(self.project_key)
+        outputs = project.get_flow().get_graph().get_successor_computables(self)
 
-        settings = self.get_settings()
-        output_refs = settings.get_flat_output_refs()
-
-        if len(output_refs) == 0:
+        if len(outputs) == 0:
             raise Exception("recipe has no outputs, can't run it")
 
-        jd = self.client.get_project(self.project_key).new_job(job_type)
-        if isinstance(settings, TrainingRecipeSettings):
-            jd.with_output(output_refs[0], object_type="SAVED_MODEL", partition=partitions)
+        first_output = outputs[0]
+
+        object_type_map = {
+            "COMPUTABLE_DATASET": "DATASET",
+            "COMPUTABLE_FOLDER": "MANAGED_FOLDER",
+            "COMPUTABLE_SAVED_MODEL": "SAVED_MODEL",
+            "COMPUTABLE_STREAMING_ENDPOINT": "STREAMING_ENDPOINT",
+        }
+        if first_output["type"] in object_type_map:
+            jd = project.new_job(job_type)
+            jd.with_output(first_output["ref"], object_type=object_type_map[first_output["type"]], partition=partitions)
         else:
-            jd.with_output(output_refs[0], partition=partitions)
+            raise Exception("recipe has unsuported output type {}, can't run it".format(first_output["type"]))
 
         if wait:
             return jd.start_and_wait()
--- a/dataikuapi/dss/recipe.py
+++ b/dataikuapi/dss/recipe.py
@@ -77,6 +77,7 @@ class DSSRecipe(object):
         :rtype: :class:`dataikuapi.dss.job.DSSJob`
         """
         project = self.client.get_project(self.project_key)
+        settings = self.get_settings()
         outputs = project.get_flow().get_graph().get_successor_computables(self)
 
         if len(outputs) == 0:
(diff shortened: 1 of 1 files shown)


instanceofme (approved):
Could probably use a more future-friendly solution with other types as well (e.g. python or custom recipe building a Folder), but still beats current behavior.

nicolasservel (commented):
This solution, while it works for saved models, seems a bit unsettling, because it would not be replicable for recipes for which we do not know the type of output (e.g. code recipe). 

I think we can build a solution that leverages the graph logic introduced recently to get the type of output. It would look like:

```python
outputs = self.client.get_project(self.project_key).get_flow().get_graph().get_successor_computables(self)

if len(outputs) == 0:
    raise Exception("recipe has no outputs, can't run it")

first_output = outputs[0]
if first_output["type"] == "COMPUTABLE_DATASET":
    first_output_object_type = "DATASET"
elif first_output["type"] == "COMPUTABLE_FOLDER":
    first_output_object_type = "MANAGED_FOLDER"
elif first_output["type"] == "COMPUTABLE_SAVED_MODEL":
    first_output_object_type = "SAVED_MODEL"
else:
    raise Exception("recipe has unsuported output type {}, can't run it".format(first_output["type"]))

jd = self.client.get_project(self.project_key).new_job(job_type)
jd.with_output(first_output["ref"], object_type=first_output_object_type, partition=partitions)
```

Of course, this needs to be extensively tested to check that it works well. 

Besides, I did not add `COMPUTABLE_STREAMING_ENDPOINT` in the example, as I don't know whether this applies to it or not.

instanceofme:
It probably does

dataikuapi/dss/recipe.py
```diff
@@ -133,6 +143,10 @@ def get_settings(self):
             return SplitRecipeSettings(self, data)
         elif type == "prepare" or type == "shaker":
             return PrepareRecipeSettings(self, data)
+        elif type == "prediction_training":
+            return TrainingRecipeSettings(self, data)
+        elif type == "clustering_training":
+            return TrainingRecipeSettings(self, data)
```
instanceofme, line 149: This (and the `TrainingRecipeSettings` definition) seems now unsused

instanceofme (approved):

dataikuapi/dss/recipe.py
```diff
@@ -76,15 +76,25 @@ def run(self, job_type="NON_RECURSIVE_FORCED_BUILD", partitions=None, wait=True,
...
 
-        jd = self.client.get_project(self.project_key).new_job(job_type)
-        jd.with_output(output_refs[0], partition=partitions)
+        first_output = outputs[0]
+
+        object_type_map = {
+            "COMPUTABLE_DATASET": "DATASET",
+            "COMPUTABLE_FOLDER": "MANAGED_FOLDER",
+            "COMPUTABLE_SAVED_MODEL": "SAVED_MODEL",
+            "COMPUTABLE_STREAMING_ENDPOINT": "STREAMING_ENDPOINT",
```
nicolasservel, line 91: So I did the test and `STREAMING_ENDPOINT` can be run, but it will produce an empty job. Instead they can be started or stopped (and we have a `dataikuapi.dss.continuousactivity.DSSContinuousActivity` for that). 

So we should:
* either fail with the generic error
* or be a little nicer and output a contextualized message saying that they should use `DSSContinuousActivity` instead

What do you think ?

nicolasservel (changes requested):

dataikuapi/dss/recipe.py
instanceofme: I have no strong opinion one way or the other 😅

mhham: After discussing with @FChataigner, it seems that we should keep the ability to run a recipe with `STREAMING_ENDPOINT` output, for the case of Sync recipes where it is possible to sync dataset -> streaming endpoint. 

For the other cases, we can fail with a contextualized message

dataikuapi/dss/recipe.py
```diff
@@ -76,15 +76,30 @@ def run(self, job_type="NON_RECURSIVE_FORCED_BUILD", partitions=None, wait=True,
         :return: the :class:`dataikuapi.dss.job.DSSJob` job handle corresponding to the built job
         :rtype: :class:`dataikuapi.dss.job.DSSJob`
         """
+        project = self.client.get_project(self.project_key)
 
-        settings = self.get_settings()
-        output_refs = settings.get_flat_output_refs()
+        continuous_recipes_names = set(activity.recipe_id for activity in project.list_continuous_activities())
```
instanceofme, line 81: Wondering how this scales for very large flows / whether it would be better to add to the backend API an indication that some recipe is continuous.
Another possibility could be to add the check on the backend side on the start call.

dataikuapi/dss/recipe.py
mhham: Revert changes related to continuous edge case to keep it simple in this PR

nicolasservel (approved):

PR #95, opened 2020-09-17
Add helpers and checks for new hyperparameter search

samronsin:
Creating the PR to collect feedback, still WIP.
I avoided creating specific classes for each algorithms, while implementing the goals we had initially:
- filter out unmodifiable settings
- catch mistakes early to avoid dreaded JSON parse error from backend
- by having setters for all dimensions
- improve discoverability of hyperparameters
- improve ease of handling hyperparameter search settings (WIP)

Could add helpers for:
- [x] `lr_settings.penalty.set_values({"l1": {"enabled": True}})` (heavy structure)
- [x] `lr_settings.multi_class.set_values("multinomial")` (admissible values check)
- [x] quick getter for dict of active algorithms

[ch53507]

TODO: 
- [x] Doctrings
- [ ] CI to assert e.g. that `save` works as expected ?

PR #108, opened 2020-12-02
Add assertions public api methods

arnaudde:
First draft of methods to be added is in the [wiki](https://analytics.dataiku.com/projects/RDWIKI/wiki/972/Feature:%20ML%20assertions#public-api-1)
Can be tested with the [dip pr](https://github.com/dataiku/dip/pull/11515) that adds tests for the methods added in this PR.

### DIFF
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -5,6 +5,7 @@ from ..utils import DataikuUTF8CSVReader
 from ..utils import DataikuStreamedHttpUTF8CSVReader
 import json, warnings
 import time
+import pprint
 from .metrics import ComputedMetrics
 from .utils import DSSDatasetSelectionBuilder, DSSFilterBuilder
 from .future import DSSFuture
@@ -711,6 +712,10 @@ class DSSMLAssertionParams(object):
     def __init__(self, data):
         self._internal_dict = data
 
+    def __repr__(self):
+        return u"DSSMLAssertionParams(\n    name= '{name}',\n    condition= {condition},\n    filter= {filter}\n)".format(
+            name=self.name, filter=pprint.pformat(self.filter).replace('\n', '\n            '), condition=self.condition)
+
     @staticmethod
     def from_params(name, a_filter, condition):
         """
@@ -782,6 +787,12 @@ class DSSMLAssertionCondition(object):
     """
     def __init__(self, data):
         self._internal_dict = data
+
+    def __repr__(self):
+        return u"DSSMLAssertionCondition(expected_valid_ratio = {}, expected_class= {}," \
+               u" expected_min={}, expected_max={})".format(
+            self.expected_valid_ratio, self.expected_class, self.expected_min, self.expected_max)
+
 
     @staticmethod
     def from_expected_class(expected_valid_ratio, expected_class):
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -628,6 +628,7 @@ class DSSMLDiagnostic(object):
     def get_type(self):
         """
         Returns the base Diagnostic type
+
         :rtype: str
         """
         return self._internal_dict["type"]
@@ -635,6 +636,7 @@ class DSSMLDiagnostic(object):
     def get_type_pretty(self):
         """
         Returns the Diagnostic type as displayed in the UI
+
         :rtype: str
         """
         return self._internal_dict["displayableType"]
@@ -642,6 +644,7 @@ class DSSMLDiagnostic(object):
     def get_message(self):
         """
         Returns the message as displayed in the UI
+
         :rtype: str
         """
         return self._internal_dict["message"]
@@ -664,6 +667,7 @@ class DSSMLAssertionsParams(object):
     def get_raw(self):
         """
         Gets the raw dictionary of the assertions parameters
+
         :rtype: dict
         """
         return self._internal_dict
@@ -739,6 +743,7 @@ class DSSMLAssertionParams(object):
     def get_raw(self):
         """
         Gets the raw dictionary of the assertion parameters
+
         :rtype: dict
         """
         return self._internal_dict
@@ -747,6 +752,7 @@ class DSSMLAssertionParams(object):
     def name(self):
         """
         Returns the assertion name
+
         :rtype: str
         """
         return self._internal_dict["name"]
@@ -759,6 +765,7 @@ class DSSMLAssertionParams(object):
     def filter(self):
         """
         Returns the assertion filter
+
         :rtype: dict
         """
         return self._internal_dict["filter"]
@@ -771,6 +778,7 @@ class DSSMLAssertionParams(object):
     def condition(self):
         """
         Returns the assertion condition
+
         :rtype: :class:`DSSMLAssertionCondition`
         """
         return DSSMLAssertionCondition(self._internal_dict["assertionCondition"])
@@ -834,6 +842,7 @@ class DSSMLAssertionCondition(object):
     def get_raw(self):
         """
         Gets the raw dictionary of the condition
+
         :rtype: dict
         """
         return self._internal_dict
@@ -843,6 +852,7 @@ class DSSMLAssertionCondition(object):
         """
         Returns the expected class or None if it is not defined. Assertion passes if the expected_valid_ratio
         of rows predicted as expected_class is attained.
+
         :rtype: str
         """
         if "expectedClass" in self._internal_dict:
@@ -859,6 +869,7 @@ class DSSMLAssertionCondition(object):
         """
         Returns the ratio of valid rows to exceed for the assertion to pass. A row is considered valid if the prediction
         is equal to the expected class for classification or in the expected range for regression.
+
         :rtype: str
         """
         return self._internal_dict["expectedValidRatio"]
@@ -872,6 +883,7 @@ class DSSMLAssertionCondition(object):
         """
         Returns the min (included) of the expected range or None if it is not defined.
         Assertion passes if the expected_valid_ratio of rows predicted between expected_min and expected_max (included) is attained.
+
         :rtype: float
         """
         if "expectedMinValue" in self._internal_dict:
@@ -888,6 +900,7 @@ class DSSMLAssertionCondition(object):
         """
         Returns the max (included) of the expected range or None if it is not defined.
         Assertion passes if the ratio of rows predicted between expected_min and expected_max (included) is attained.
+
         :rtype: float
         """
         if "expectedMaxValue" in self._internal_dict:
@@ -911,6 +924,7 @@ class DSSMLAssertionsMetrics(object):
     def get_raw(self):
         """
         Gets the raw dictionary of the assertions metrics
+
         :rtype: dict
         """
         return self._internal_dict
@@ -934,6 +948,7 @@ class DSSMLAssertionsMetrics(object):
     def passing_assertions_ratio(self):
         """
         Returns the ratio of passing assertions
+
         :rtype: float
         """
         return self._internal_dict['passingAssertionsRatio']
@@ -954,6 +969,7 @@ class DSSMLAssertionMetric(object):
     def get_raw(self):
         """
         Gets the raw dictionary of metrics of one assertion
+
         :rtype: dict
         """
         return self._internal_dict
@@ -962,6 +978,7 @@ class DSSMLAssertionMetric(object):
     def name(self):
         """
         Returns the assertion name
+
         :rtype: str
         """
         return self._internal_dict["name"]
@@ -970,6 +987,7 @@ class DSSMLAssertionMetric(object):
     def result(self):
         """
         Returns whether the assertion passes
+
         :rtype: bool
         """
         return self._internal_dict["result"]
@@ -979,6 +997,7 @@ class DSSMLAssertionMetric(object):
         """
         Returns the ratio of rows in the assertion population with prediction equals to the expected class
         for classification or in the expected range for regression
+
         :rtype: float
         """
         return self._internal_dict["validRatio"]
@@ -987,6 +1006,7 @@ class DSSMLAssertionMetric(object):
     def nb_matching_rows(self):
         """
         Returns the number of rows matching filter
+
         :rtype: int
         """
         return self._internal_dict["nbMatchingRows"]
@@ -995,6 +1015,7 @@ class DSSMLAssertionMetric(object):
     def nb_dropped_rows(self):
         """
         Returns the number of rows dropped by the model's preprocessing
+
         :rtype: int
         """
         return self._internal_dict["nbDroppedRows"]
@@ -1161,6 +1182,7 @@ class DSSTrainedPredictionModelDetails(DSSTrainedModelDetails):
     def get_assertions_metrics(self):
         """
         Retrieves assertions metrics computed for this trained model
+
         :returns: an object representing assertion metrics
         :rtype: :class:`DSSMLAssertionsMetrics`
         """
@@ -1858,6 +1880,7 @@ class DSSTrainedClusteringModelDetails(DSSTrainedModelDetails):
     def get_performance_metrics(self):
         """
         Returns all performance metrics for this clustering model.
+
         :returns: a dict of performance metrics values
         :rtype: dict
         """
@@ -1890,6 +1913,7 @@ class DSSTrainedClusteringModelDetails(DSSTrainedModelDetails):
     def get_actual_modeling_params(self):
         """
         Gets the actual / resolved parameters that were used to train this model.
+
         :return: A dictionary, which contains at least a "resolved" key
         :rtype: dict
         """
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -731,7 +731,8 @@ class DSSMLAssertionParams(object):
         Creates assertion parameters from name, filter and condition
 
         :param str name: Name of the assertion
-        :param object a_filter: A dict representing the filter to select assertion population
+        :param dict a_filter: A dict representing the filter to select assertion population. You can use
+        a :class:`~dataikuapi.dss.utils.DSSFilterBuilder` to build the settings of the filter
         :param object condition: A :class:`DSSMLAssertionCondition` for the assertion to be successful
 
         :rtype: :class:`DSSMLAssertionParams`
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -5,7 +5,6 @@ from ..utils import DataikuUTF8CSVReader
 from ..utils import DataikuStreamedHttpUTF8CSVReader
 import json, warnings
 import time
-import pprint
 from .metrics import ComputedMetrics
 from .utils import DSSDatasetSelectionBuilder, DSSFilterBuilder
 from .future import DSSFuture
@@ -720,8 +719,8 @@ class DSSMLAssertionParams(object):
         self._internal_dict = data
 
     def __repr__(self):
-        return u"DSSMLAssertionParams(\n    name= '{name}',\n    condition= {condition},\n    filter= {filter}\n)".format(
-            name=self.name, filter=pprint.pformat(self.filter).replace('\n', '\n            '), condition=self.condition)
+        return u"DSSMLAssertionParams(name= '{name}', condition= {condition}, filter= {filter})".format(
+            name=self.name, filter=self.filter, condition=self.condition)
 
     @staticmethod
     def from_params(name, a_filter, condition):
@@ -963,8 +962,8 @@ class DSSMLAssertionMetric(object):
         self._internal_dict = data
 
     def __repr__(self):
-        return u"DSSMLAssertionParams(\n    name='{}',\n    result={},\n    valid_ratio={},\n    nb_matching_rows={}," \
-               u"\n    nb_dropped_rows={}\n)".format(self.name, self.result, self.valid_ratio, self.nb_matching_rows,
+        return u"DSSMLAssertionParams(name='{}', result={}, valid_ratio={}, nb_matching_rows={}," \
+               u" nb_dropped_rows={})".format(self.name, self.result, self.valid_ratio, self.nb_matching_rows,
                                                  self.nb_dropped_rows)
     def get_raw(self):
         """
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -928,19 +928,19 @@ class DSSMLAssertionsMetrics(object):
         """
         return self._internal_dict
 
-    def get_metric(self, assertion_name):
+    def get_metrics(self, assertion_name):
         """
-        Retrieves the metric computed for this trained model for the assertion with the provided name (or None if no
+        Retrieves the metrics computed for this trained model for the assertion with the provided name (or None if no
         assertion with that name exists)
 
         :param str assertion_name: Name of the assertion
 
         :returns: an object representing assertion metrics or None if no assertion with that name exists
-        :rtype: :class:`DSSMLAssertionMetric`
+        :rtype: :class:`DSSMLAssertionMetrics`
         """
-        for assertion_metric_dict in self._internal_dict["perAssertion"]:
-            if assertion_name == assertion_metric_dict["name"]:
-                return DSSMLAssertionMetric(assertion_metric_dict)
+        for assertion_metrics_dict in self._internal_dict["perAssertion"]:
+            if assertion_name == assertion_metrics_dict["name"]:
+                return DSSMLAssertionMetrics(assertion_metrics_dict)
         return None
 
     @property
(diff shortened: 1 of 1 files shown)


dataikuapi/dss/ml.py
```diff
@@ -643,6 +660,335 @@ def __repr__(self):
...
+        """
+        if "expectedClass" in self._internal_dict:
+            return self._internal_dict["expectedClass"]
+        else:
+            return None
+
+    @expected_class.setter
+    def expected_class(self, expected_class):
+        if self.expected_range is not None:
+            raise ValueError("Expected class and expected range can't be both set")
```
instanceofme, line 860: - Should be rather enforced by the backend (on save)
- Arguably if we do want/accept the tight coupling, this method could rather unset the range + set the class. (I don't think we do want that though, as it seems to me there is no use case)

dataikuapi/dss/ml.py
```diff
@@ -643,6 +660,335 @@ def __repr__(self):
...
+    @expected_valid_ratio.setter
+    def expected_valid_ratio(self, expected_valid_ratio):
+        self._internal_dict["successRatio"] = expected_valid_ratio
+
+    @property
+    def expected_range(self):
+        """
+        Returns the expected range on which the valid ratio will be calculated
+        :rtype: tuple(float,float)
+        """
```
instanceofme, line 880: To check: consistency with #95 where we expose a range as an object with min and max.

dataikuapi/dss/ml.py
```diff
@@ -643,6 +660,335 @@ def __repr__(self):
...
+        if expected_range[0] > expected_range:
+            raise ValueError("Expected range needs to be sorted in ascending order. (min value, max value)")
+        self._internal_dict["expectedMinValue"] = expected_range[0]
+        self._internal_dict["expectedMaxValue"] = expected_range[1]
+
+
+class DSSMLAssertionsMetrics(object):
+    """
+    Object that represents the per assertion metrics for all assertions on a trained model
+    Do not create this object directly, use :meth:`dataikuapi.dss.ml.DSSTrainedPredictionModelDetails.get_per_assertion_metrics()` instead
```
instanceofme, line 901: The method is called `get_assertions_metrics` no? (which i think is better)

dataikuapi/dss/ml.py
```diff
@@ -643,6 +660,335 @@ def __repr__(self):
...
+        Returns the ratio of passing assertions
+        :rtype: float
+        """
+        return self._internal_dict['positiveAssertionsRatio']
+
+
+class DSSMLAssertionMetric(object):
+    """
+    Object that represents the result of an assertion on a trained model
+    Do not create this object directly, use :meth:`dataikuapi.dss.ml.DSSMLPerAssertionMetrics.get_assertion_metric(self, assertion_name)` instead
```
instanceofme, line 939: The class is called `DSSMLAssertionsMetrics` (no `Per`). To check: doc method/class names everywhere in new docstrings.

dataikuapi/dss/ml.py
```diff
@@ -643,6 +660,335 @@ def __repr__(self):
...
+        self._internal_dict = data
+
+    def get_raw(self):
+        """
+        Gets the raw dictionary of the assertions metrics
+        :rtype: dict
+        """
+        return self._internal_dict
+
+    def get_per_assertion_metric(self, assertion_name):
```
instanceofme, line 913: Either `get_assertion_metric` (or even `get_metric`?)

dataikuapi/dss/ml.py
```diff
@@ -643,6 +660,335 @@ def __repr__(self):
...
+        """
+        Retrieves the metric computed for this trained model for the assertion with the provided name (or None)
```
instanceofme, line 915: Specify in which case `None`, even when it seems obvious to you

dataikuapi/dss/ml.py
```diff
@@ -643,6 +660,335 @@ def __repr__(self):
...
+
+        :param str assertion_name: Name of the assertion
+
+        :returns: an object representing assertion metrics
```
instanceofme, line 919: … or None if …

dataikuapi/dss/ml.py
```diff
@@ -643,6 +660,335 @@ def __repr__(self):
...
+        """
+        Returns whether the assertion pass
+        :rtype: bool
+        """
+        return self._internal_dict["result"]
+
+    @property
+    def valid_ratio(self):
+        """
+        Returns the ratio of passing rows in the assertion population
```
instanceofme, line 970: Not sure we want to use "passing" as we said it's more at the assertion level. "…of rows with prediction of the expected class or in the expected range"? Admittedly a mouthful.

dataikuapi/dss/ml.py
```diff
@@ -643,6 +660,335 @@ def __repr__(self):
...
+        """
+        Returns the number of rows matching filter
+        :rtype: int
+        """
+        return self._internal_dict["nbMatchingRows"]
+
+    @property
+    def nb_dropped_rows(self):
+        """
+        Returns the number of rows dropped by the preprocessing
```
instanceofme, line 986:
```suggestion
        Returns the number of rows dropped by the model's preprocessing
```

dataikuapi/dss/ml.py
```diff
@@ -801,6 +1147,22 @@ def get_performance_metrics(self):
...
         return clean_snippet
 
+    @property
+    def assertions_metrics(self):
+        """
+        Retrieves assertions metrics computed for this trained model
+        :returns: an object representing assertion metrics
+        :rtype: :class:`dataikuapi.dss.ml.DSSMLAssertionsMetrics`
+        """
+        return self.get_assertions_metrics()
```
instanceofme, line 1157: Not sure a property helps if you have a no-argument getter already. Seems like just an additional thing to maintain. (Saw it elsewhere too).

instanceofme (changes requested):

dataikuapi/dss/ml.py
arnaudde: I find it a bit nicer to write `model_details.assertions_metrics.get_per_assertion_metric("women")` than `model_details.get_assertions_metrics().get_per_assertion_metric("women")`. Chaining getters seems uglier. But I am happy to change if it is not worth the maintenance cost

instanceofme: I don't think it is indeed. I'm ok on "terminal" objects to easily get/set a direct value, but everywhere else where we get complex/custom objects, we have getters…

dataikuapi/dss/ml.py
```diff
@@ -643,6 +660,335 @@ def __repr__(self):
...
 
 
+class DSSMLAssertionsParams(object):
+    """
+    Object that represents parameters for all assertions of a ml task
+    Do not create this object directly, use :meth:`DSSPredictionMLTaskSettings.get_assertions_params()` instead
+    """
+
+    @staticmethod
+    def check_assertion_names_are_uniq(assertion_params_list):
```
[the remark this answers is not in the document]
arnaudde: I assumed that the uniqueness of assertion names would stay for long. But I'll delete it to avoid coupling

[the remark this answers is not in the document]
arnaudde: Done in 15343ea

dataikuapi/dss/ml.py
```diff
@@ -643,6 +660,335 @@ def __repr__(self):
...
+                    assertion_dict['name']))
+            _[assertion_dict['name']] = True
+
+    def __init__(self, data):
+        self._internal_dict = data
+        self.check_assertion_names_are_uniq(data["assertions"])
+
+    def get_raw(self):
+        """
+        Gets the raw dictionary of the assertions params
```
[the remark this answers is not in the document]
arnaudde: Done only for the new code ("params" is present multiple times in older docstring) in 15343ea

dataikuapi/dss/ml.py
```diff
@@ -643,6 +660,335 @@ def __repr__(self):
...
+    """
+    Object that represents parameters for one assertion
+    Do not create this object directly, use :meth:`dataikuapi.dss.ml.DSSMLAssertionsParams.get_assertion(assertion_name)` or
+    `create_from_parts(name, a_filter, condition)` instead
+    """
+    def __init__(self, data):
+        self._internal_dict = data
+
+    @staticmethod
+    def create_from_parts(name, a_filter, condition):
```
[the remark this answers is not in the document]
arnaudde: `from` is not a valide identifier renaming to `from_parts`. Done in 5e72bd5

dataikuapi/dss/ml.py
```diff
@@ -643,6 +660,335 @@ def __repr__(self):
...
+    """
+      Object that represents an assertion condition
+      Do not create this object directly, use :meth:`dataikuapi.dss.ml.DSSMLAssertionParams.condition`, :meth:`dataikuapi.dss.ml.DSSMLAssertionCondition.create_from_expected_class(expected_valid_ratio, expected_class)`
+      or :meth:`dataikuapi.dss.ml.DSSMLAssertionCondition.create_from_expected_range(expected_valid_ratio, expected_range)` instead
+    """
+    def __init__(self, data):
+        self._internal_dict = data
+
+    @staticmethod
+    def create_from_expected_class(expected_valid_ratio, expected_class):
```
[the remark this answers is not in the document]
arnaudde: Done in 5e72bd5

dataikuapi/dss/ml.py
```diff
@@ -643,6 +660,335 @@ def __repr__(self):
...
+        return assertion_condition
+
+    @staticmethod
+    def create_from_expected_range(expected_valid_ratio, expected_range):
+        """
+        Creates an assertion condition from an expected valid ratio and an expected range
+
+        :param float expected_valid_ratio: Ratio of valid rows to exceed for the assertion to pass
+        :param tuple(float,float) expected_range: Range of values (min, max) where the prediction will be considered as
+        valid for the `expected_valid_ratio`
```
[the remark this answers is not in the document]
arnaudde: Done in 9ede2b4

dataikuapi/dss/ml.py
```diff
@@ -792,7 +781,7 @@ def condition(self):
     @condition.setter
     def condition(self, condition):
         if not isinstance(condition, DSSMLAssertionCondition):
-            raise ValueError('Wrong type for assertion condition: {}.format(type(condition)))
+            raise ValueError('Wrong type for assertion condition: {}.format(type(condition))')
```
instanceofme, line 784: Does not look right :)
```suggestion
            raise ValueError('Wrong type for assertion condition: {}'.format(type(condition)))
```

dataikuapi/dss/ml.py
```diff
@@ -643,6 +660,335 @@ def __repr__(self):
...
+
+        :rtype: :class:`dataikuapi.dss.ml.DSSMLAssertionCondition`
+        """
+        assertion_condition = DSSMLAssertionCondition({})
+        assertion_condition.expected_valid_ratio = expected_valid_ratio
+        assertion_condition.expected_class = expected_class
+        return assertion_condition
+
+    @staticmethod
+    def create_from_expected_range(expected_valid_ratio, expected_range):
```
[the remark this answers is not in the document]
arnaudde: Done in 9ede2b4

dataikuapi/dss/ml.py
arnaudde: Thanks !

dataikuapi/dss/ml.py
```diff
@@ -810,19 +800,22 @@ def create_from_expected_class(expected_valid_ratio, expected_class):
         return assertion_condition
 
     @staticmethod
-    def create_from_expected_range(expected_valid_ratio, expected_range):
+    def from_expected_range(expected_valid_ratio, expected_min, expected_max):
         """
-        Creates an assertion condition from an expected valid ratio and an expected range
+        Creates an assertion condition from expected valid ratio and range. The expected range is the
+        interval between expected_min and expected_max where the predictions and therefore the rows will be considered
+        valid.
```
instanceofme, line 807:
```suggestion
        Creates an assertion condition from expected valid ratio and range.
        The expected range is the interval between expected_min and expected_max (included)
        for the predictions in which the rows will be considered valid.
```

dataikuapi/dss/ml.py
```diff
@@ -835,7 +828,8 @@ def get_raw(self):
     @property
     def expected_class(self):
         """
-        Returns the expected class on which the valid ratio will be calculated
+        Returns the expected class or None if it is not defined. Assertion passes if the ratio of rows predicted
+        as expected_class is attained
```
instanceofme, line 832:
```suggestion
        Returns the expected class or None if it is not defined. Assertion passes if the expected_valid_ratio
        of rows predicted as expected_class is attained.
```

dataikuapi/dss/ml.py
```diff
@@ -845,14 +839,13 @@ def expected_class(self):
...
-        if self.expected_range is not None:
-            raise ValueError("Expected class and expected range can't be both set")
         self._internal_dict["expectedClass"] = expected_class
 
     @property
     def expected_valid_ratio(self):
         """
-        Returns the ratio of valid rows to exceed for the assertion to pass
+        Returns the ratio of valid rows to exceed for the assertion to pass. A row is considered valid if the prediction
+        is equal to the `expected_class` for classification or in the expected range for regresion
```
instanceofme, line 848:
```suggestion
        is equal to the `expected_class` for classification or in the expected range for regression
```

dataikuapi/dss/ml.py
```diff
@@ -862,32 +855,42 @@ def expected_valid_ratio(self, expected_valid_ratio):
         self._internal_dict["successRatio"] = expected_valid_ratio
 
     @property
-    def expected_range(self):
+    def expected_min(self):
+        """
+        Returns the min (included) of the expected range or None if it is not defined. Assertion passes if the ratio of rows predicted
+        between expected_min and expected_max is attained
```
instanceofme, line 861:
```suggestion
        Returns the min (included) of the expected range or None if it is not defined.
        Assertion passes if the ratio of rows predicted between expected_min and expected_max is attained.
```

dataikuapi/dss/ml.py
```diff
@@ -862,32 +855,42 @@ def expected_valid_ratio(self, expected_valid_ratio):
...
+    def expected_min(self, expected_min):
+        self._internal_dict["expectedMinValue"] = expected_min
+
+    @property
+    def expected_max(self):
         """
-        Returns the expected range on which the valid ratio will be calculated
-        :rtype: tuple(float,float)
+        Returns the max (included) of the expected range or None if it is not defined. Assertion passes if the ratio of rows predicted
+        between expected_min and expected_max is attained
```
instanceofme, line 877:
```suggestion
        Returns the max (included) of the expected range or None if it is not defined.
        Assertion passes if the ratio of rows predicted between expected_min and expected_max is attained.
```

instanceofme (commented):
Looks better. A few wording / readability nitpicks.

arnaudde:
Public api ml task tests and the ones added in the [twin dip pr](https://github.com/dataiku/dip/pull/11515) are passsing ✅

nicolasservel (changes requested):
* You use sometimes ``:class:`dataikuapi.dss.ml.DSSMLAssertionMetric` `` without `~` or ``:class:`~dataikuapi.dss.utils.DSSFilter` `` for referencing classes, is there a specific reason or should it be harmonized ?

* Most of the introduced classes could use string representation, to ease the use in notebooks

arnaudde:
As fred T explained `~` in class reference are for sphynx syntax display purposes.
> https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html
> If you prefix the content with ~, the link text will only be the last component of the target. For example, :py:meth:`~Queue.Queue.get` will refer to Queue.Queue.get but only display get as the link text. This does not work with all cross-reference roles, but is domain specific.

dataikuapi/dss/ml.py
```diff
@@ -643,6 +652,351 @@ def __repr__(self):
...
+    Object that represents parameters for one assertion
+    Do not create this object directly, use :meth:`DSSMLAssertionsParams.get_assertion(assertion_name)` or
+    `from_parts(name, a_filter, condition)` instead
+    """
+    def __init__(self, data):
+        self._internal_dict = data
+
+    def __repr__(self):
+        return u"DSSMLAssertionParams(\n    name= '{name}',\n    condition= {condition},\n    filter= {filter}\n)".format(
+            name=self.name, filter=pprint.pformat(self.filter).replace('\n', '\n            '), condition=self.condition)
```
instanceofme, line 717: Not certain we want multiline repr, as it makes it harder to grep in logs. IMO acceptable to have `DSSMLAssertionParams(<internal dict repr>)`

dataikuapi/dss/ml.py
```diff
@@ -643,6 +652,351 @@ def __repr__(self):
...
+        self._internal_dict = data
+
+    def get_raw(self):
+        """
+        Gets the raw dictionary of the assertions metrics
+        :rtype: dict
+        """
+        return self._internal_dict
+
+    def get_metric(self, assertion_name):
```
instanceofme, line 915: Open question: `metric` or `metrics`? `metric` makes it easier / more obvious to distinguish per assertion vs all, but technically we return several metrics for each assertion…
We could also dodge the question and call it `get_result` / `DSSAssertionResult` 😄 (in which case its `result` property could be renamed `passes`).

arnaudde:
Methods are working (no failure, result looks good) on a python3.6 notebook

instanceofme (approved):
Essentially LGTM

dataikuapi/dss/ml.py
arnaudde: The question of `metric` vs `result` is impacting the variable name consistency in dip as well. IMO, we should decide which is better and use only one everywhere

nicolasservel (changes requested):

dataikuapi/dss/ml.py
nicolasservel: I have added my proposal in my latest review. For me it should be `DSSMLAssertionMetrics` & `get_assertion_metrics`, what do you think ?

instanceofme: - Agree with `DSSMLAssertionMetrics` and the `_metrics` plural for a single assertion
- Arnaud initially named the method `get_assertion_metrics`, and I suggested `get_metrics` as to me it seemed redundant (`trained.get_assertions_metrics().get_assertion_metric("name")` → `trained.get_assertions_metrics().get_metric("name")`). However that opinion is loosely held, so if you both think the clarity makes it worth the redundancy, I'll yield.

dataikuapi/dss/ml.py
```diff
@@ -643,6 +652,351 @@ def __repr__(self):
...
+        """
+        for assertion_dict in self._internal_dict["assertions"]:
+            if assertion_dict["name"] == assertion_name:
+                return DSSMLAssertionParams(assertion_dict)
+        return None
+
+    def add_assertion(self, assertion_params):
+        """
+        Adds parameters of an assertion to the assertions parameters of the ml task.
+        :param object  assertion_params: A :class:`DSSMLAssertionParams` representing parameters of the assertion
```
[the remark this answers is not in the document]
arnaudde: I don't understand when we should use `object` and when we should use the `DSS*class`. I mainly see usage of object after the `:param` keyword

dataikuapi/dss/ml.py
```diff
@@ -643,6 +652,351 @@ def __repr__(self):
...
+    def filter(self):
+        """
+        Returns the assertion filter
+        :rtype: dict
+        """
+        return self._internal_dict["filter"]
+
+    @filter.setter
+    def filter(self, filter):
+        self._internal_dict["filter"] = filter
```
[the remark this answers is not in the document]
arnaudde: filter is a protected python keyword so the parameter can't be called filter

dataikuapi/dss/ml.py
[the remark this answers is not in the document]
nicolasservel: The docstring should be the closest to what we expect from the function, and is a poor replacement of static typing for python. It allows for example, when editing code in an IDE, to have a smarter autocomplete. `object` is used when we don't know the type of the corresponding parameter

dataikuapi/dss/ml.py
nicolasservel: You convinced me with your example, let's go with `get_metrics` :-)

dataikuapi/dss/ml.py
```diff
@@ -643,6 +652,351 @@ def __repr__(self):
                                                           msg=self._internal_dict["message"])
 
 
+class DSSMLAssertionsParams(object):
```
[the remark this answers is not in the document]
arnaudde: How would you expect this string representation to be ? I did not add it because I am afraid it would be too verbose if we print every assertionParam string representation of the `perAssertion` attribute.
Should we just print something like this ?
```python
DSSMLAssertionsParams('assertions': [ DSSMLAssertionParams('Assertion1',...), DSSMLAssertionParams('Assertion2',...), DSSMLAssertionParams('Assertion3',...), ...])
```

dataikuapi/dss/ml.py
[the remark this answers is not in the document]
arnaudde: As discussed, because class names can become long, we will stick with the following notation in the doc
```
:param object object_name: A :class:`DSSMLClass_name` representing ...
```

Screen Shot 2020-12-16 at 14 44 11

dataikuapi/dss/ml.py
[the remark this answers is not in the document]
nicolasservel: Yes, or even just `DSSMLAssertionsParams(... internal_dict ...)` for simplicity, I see that as a helper when building scripts in a notebook.

dataikuapi/dss/ml.py
```diff
@@ -643,6 +654,386 @@ def __repr__(self):
...
+
+        :rtype: float
+        """
+        return self._internal_dict['passingAssertionsRatio']
+
+
+class DSSMLAssertionMetrics(object):
+    """
+    Object that represents the result of an assertion on a trained model
+    Do not create this object directly, use :meth:`DSSMLAssertionMetrics.get_metrics(self, assertion_name)` instead
```
nicolasservel, line 972: This does not render well when put to the documentation

image


```suggestion
    Do not create this object directly, use :meth:`DSSMLAssertionMetrics.get_metrics` instead
```

dataikuapi/dss/ml.py
```diff
@@ -643,6 +654,386 @@ def __repr__(self):
                                                           msg=self._internal_dict["message"])
 
 
+class DSSMLAssertionsParams(object):
+    """
+    Object that represents parameters for all assertions of a ml task
+    Do not create this object directly, use :meth:`DSSPredictionMLTaskSettings.get_assertions_params()` instead
```
nicolasservel, line 660:
```suggestion
    Do not create this object directly, use :meth:`DSSPredictionMLTaskSettings.get_assertions_params` instead
```

dataikuapi/dss/ml.py
```diff
@@ -643,6 +654,386 @@ def __repr__(self):
...
+                del self._internal_dict["assertions"][idx]
+                return
+        raise ValueError('No assertion found with name: {}'.format(assertion_name))
+
+
+class DSSMLAssertionParams(object):
+    """
+    Object that represents parameters for one assertion
+    Do not create this object directly, use :meth:`DSSMLAssertionsParams.get_assertion(assertion_name)` or
+    `from_params(name, a_filter, condition)` instead
```
nicolasservel, line 728:
```suggestion
    Do not create this object directly, use :meth:`DSSMLAssertionsParams.get_assertion` or
    :meth:`from_params` instead
```

dataikuapi/dss/ml.py
```diff
@@ -643,6 +654,386 @@ def __repr__(self):
...
+    def condition(self, condition):
+        if not isinstance(condition, DSSMLAssertionCondition):
+            raise ValueError('Wrong type for assertion condition: {}'.format(type(condition)))
+        self._internal_dict["assertionCondition"] = condition.get_raw()
+
+
+class DSSMLAssertionCondition(object):
+    """
+    Object that represents an assertion condition
+    Do not create this object directly, use :meth:`DSSMLAssertionParams.condition`, :meth:`DSSMLAssertionCondition.from_expected_class(expected_valid_ratio, expected_class)`
```
nicolasservel, line 807:
```suggestion
    Do not create this object directly, use :meth:`DSSMLAssertionParams.condition`, :meth:`DSSMLAssertionCondition.from_expected_class`
```

dataikuapi/dss/ml.py
```diff
@@ -643,6 +654,386 @@ def __repr__(self):
...
+    or :meth:`DSSMLAssertionCondition.from_expected_range(expected_valid_ratio, expected_min, expected_max)` instead
```
nicolasservel, line 808:
```suggestion
    or :meth:`DSSMLAssertionCondition.from_expected_range` instead
```

dataikuapi/dss/ml.py
```diff
@@ -643,6 +654,386 @@ def __repr__(self):
...
+
+    @expected_max.setter
+    def expected_max(self, expected_max):
+        self._internal_dict["expectedMaxValue"] = expected_max
+
+
+class DSSMLAssertionsMetrics(object):
+    """
+    Object that represents the assertions metrics for all assertions on a trained model
+    Do not create this object directly, use :meth:`DSSTrainedPredictionModelDetails.get_assertions_metrics()` instead
```
nicolasservel, line 928:
```suggestion
    Do not create this object directly, use :meth:`DSSTrainedPredictionModelDetails.get_assertions_metrics` instead
```

nicolasservel (changes requested):
Last nitpicks on formatting docstring, otherwise the API works smoothly :-)

nicolasservel (approved):

PR #119, opened 2021-01-22
Add syntactic sugar for numerical hyperparameters

samronsin:
[CH58941](https://app.clubhouse.io/dataiku/story/58941/public-api-python-client-add-syntactic-sugar-for-numericalhyperparametersettings)

### DIFF
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -743,7 +743,7 @@ class NumericalHyperparameterSettings(HyperparameterSettings):
         - the definition mode of the current numerical hyperparameter to "EXPLICIT"
 
         :param values: the explicit list of numerical values considered for this hyperparameter in the search
-        :type values: list of float | int
+        :type values: list of float | list of int
         """
         self.values = values
         self.definition_mode = "EXPLICIT"
@@ -760,7 +760,7 @@ class NumericalHyperparameterSettings(HyperparameterSettings):
     def values(self, values):
         """
         :param values: the explicit list of numerical values considered for this hyperparameter in the search
-        :type values: list of float | int
+        :type values: list of float | list of int
         """
         error_message = "Invalid values input type for hyperparameter " \
                         "\"{}\": ".format(self.name) + \
@@ -1064,12 +1064,19 @@ class PredictionAlgorithmSettings(dict):
                     target.set_values(value)
                 elif isinstance(target, NumericalHyperparameterSettings):
                     if isinstance(value, list):
-                        target.set_explicit_values(values=value)
+                        # algo.hyperparam = [x, y, z]
+                        target.set_explicit_values(values=1)
                     elif isinstance(value, Range):
+                        # algo.hyperparam = Range(min=x, max=y, nb_values=z)
                         target.set_range(min=value.min, max=value.max, nb_values=value.nb_values)
+                    elif isinstance(value, NumericalHyperparameterSettings):
+                        # algo.hyperparam = other_algo.other_hyperparam
+                        target.set_range(min=value.range.min, max=value.range.max, nb_values=value.range.nb_values)
+                        target.set_explicit_values(values=value.values.copy())
+                        target.definition_mode = value.definition_mode
                     else:
                         raise TypeError(("Invalid type for NumericalHyperparameterSettings {}\n" +
-                                        "Expecting either a list or a Range").format(attr_name))
+                                        "Expecting either list, Range or NumericalHyperparameterSettings").format(attr_name))
                 else:
                     # simple parameter
                     assert isinstance(value, type(target)), "Invalid type {} for parameter {}: expected {}".format(type(value), attr_name, type(target))
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -812,6 +812,10 @@ class NumericalHyperparameterSettings(HyperparameterSettings):
                 self._algo_settings[self.name]["range"]["nbValues"] = nb_values
 
     class RangeSettings(object):
+        """
+        Class designed to represent the range of a numerical hyperparameter by pointing to the algorithm settings
+        Should not be used directly by end users of the API
+        """
 
         def __init__(self, numerical_hyperparameter_settings):
             self._numerical_hyperparameter_settings = numerical_hyperparameter_settings
@@ -890,6 +894,10 @@ class NumericalHyperparameterSettings(HyperparameterSettings):
 
 
 class Range(object):
+    """
+    Class designed to represent the range of a numerical hyperparameter by storing its attributes (min, max, nb_values)
+    End users should use this class to define explicitly the parameters of the range of a numerical hyperparameter
+    """
 
     def _check_input(self, value):
         assert isinstance(value, (int, float)), "Invalid input type for Range: {}".format(type(value))
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -811,6 +811,63 @@ class NumericalHyperparameterSettings(HyperparameterSettings):
             if nb_values is not None:
                 self._algo_settings[self.name]["range"]["nbValues"] = nb_values
 
+    class Range(object):
+
+        def __init__(self, numerical_hyperparameter_settings):
+            self._numerical_hyperparameter_settings = numerical_hyperparameter_settings
+            self._range_dict = self._numerical_hyperparameter_settings._algo_settings[numerical_hyperparameter_settings.name]["range"]
+
+        def __repr__(self):
+            return "Range(min={}, max={}, nb_values={})".format(self.min, self.max, self.nb_values)
+
+        @property
+        def min(self):
+            """
+            :return: the lower bound of the Range for this hyperparameter
+            :rtype: float | int
+            """
+            return self._range_dict["min"]
+
+        @min.setter
+        def min(self, value):
+            """
+            :param value: the lower bound of the Range this hyperparameter
+            :type value: float | int
+            """
+            self._numerical_hyperparameter_settings._set_range(min=value)
+
+        @property
+        def max(self):
+            """
+            :return: the upper bound of the Range this hyperparameter
+            :rtype: float | int
+            """
+            return self._range_dict["max"]
+
+        @max.setter
+        def max(self, value):
+            """
+            :param value: the upper bound of the Range for this hyperparameter
+            :type value: float | int
+            """
+            self._numerical_hyperparameter_settings._set_range(max=value)
+
+        @property
+        def nb_values(self):
+            """
+            :return: for grid-search ("GRID" strategy) only, the number of values between min and max to consider
+            :rtype: int
+            """
+            return self._range_dict["nbValues"]
+
+        @nb_values.setter
+        def nb_values(self, value):
+            """
+            :param value: for grid-search ("GRID" strategy) only, the number of values between min and max to consider
+            :type value: int
+            """
+            self._numerical_hyperparameter_settings._set_range(nb_values=value)
+
     def set_range(self, min=None, max=None, nb_values=None):
         """
         Sets both:
@@ -829,7 +886,7 @@ class NumericalHyperparameterSettings(HyperparameterSettings):
 
     @property
     def range(self):
-        return RangeSettings(self)
+        return NumericalHyperparameterSettings.Range(self)
 
 
 class Range(object):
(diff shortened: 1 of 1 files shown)


dataikuapi/dss/ml.py
```diff
@@ -829,17 +829,28 @@ def set_range(self, min=None, max=None, nb_values=None):
 
     @property
     def range(self):
```
mhham, line 831: Not sure about this, but maybe rename it to `range_settings` to be sure we do not confuse `Range` and `RangeSettings`

dataikuapi/dss/ml.py
```diff
@@ -829,17 +829,28 @@ def set_range(self, min=None, max=None, nb_values=None):
...
-        return Range(self)
+        return RangeSettings(self)
 
 
 class Range(object):
```
mhham, line 836: Add a docstring here and for `RangeSettings` to be sure the user understands the difference ?

dataikuapi/dss/ml.py
```diff
@@ -829,17 +829,28 @@ def set_range(self, min=None, max=None, nb_values=None):
...
+    def __init__(self, min, max, nb_values=3):
+        self.min = min
+        self.max = max
+        self.nb_values = nb_values
+
+    def __repr__(self):
+        return "Range(min={}, max={}, nb_values={})".format(self.min, self.max, self.nb_values)
+
+
+class RangeSettings(object):
```
mhham, line 846: The main idea is that the user should not instanciate the `RangeSettings` class himself. Maybe nest this class inside `NumericalHyperparameterSettings` and call it using `self.RangeSettings` , to make it very clear that it has no purpose outside of `NumericalHyperparameterSettings`?

samronsin: Nesting sounds like the best solution to "hide" the `RangeSettings` class indeed, thanks for the suggestion !

dataikuapi/dss/ml.py
```diff
@@ -811,6 +811,63 @@ def _set_range(self, min=None, max=None, nb_values=None):
...
                 self._algo_settings[self.name]["range"]["nbValues"] = nb_values
 
+    class Range(object):
+
+        def __init__(self, numerical_hyperparameter_settings):
+            self._numerical_hyperparameter_settings = numerical_hyperparameter_settings
+            self._range_dict = self._numerical_hyperparameter_settings._algo_settings[numerical_hyperparameter_settings.name]["range"]
+
+        def __repr__(self):
+            return "Range(min={}, max={}, nb_values={})".format(self.min, self.max, self.nb_values)
```
mhham, line 821: To be a bit more explicit
```suggestion
            return "NumericalHyperparameterSettings.Range(min={}, max={}, nb_values={})".format(self.min, self.max, self.nb_values)
```

mhham (approved):

mhham:
Looks good, maybe add docstrings and/or document the syntactic sugar

dataikuapi/dss/ml.py
```diff
@@ -1047,7 +1057,10 @@ def __setattr__(self, attr_name, value):
                 elif isinstance(target, CategoricalHyperparameterSettings):
                     target.set_values(value)
                 elif isinstance(target, NumericalHyperparameterSettings):
```
nicolasservel, line 1059: Maybe overthinking this, but would be interesting to be able to pass an instance of `NumericalHyperparameterSettings.Range`. I can imagine copying parameters from one algorithm to another for a particular hyperparameter.

nicolasservel (changes requested):

dataikuapi/dss/ml.py
samronsin: I don't see the pro of renaming this property. To me the point is to still be able to write e.g. `rf_settings.n_estimators.range.min = 10` and also e.g. `rf_settings.n_estimators = Range(10, 100)`.

dataikuapi/dss/ml.py
samronsin: Hum. I didn't think of that but that makes sense !

dataikuapi/dss/ml.py
```diff
@@ -811,6 +811,63 @@ def _set_range(self, min=None, max=None, nb_values=None):
             if nb_values is not None:
                 self._algo_settings[self.name]["range"]["nbValues"] = nb_values
 
+    class Range(object):
```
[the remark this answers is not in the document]
samronsin: OK, yes sure I've been circling back and forth between `Range` and `RangeSettings` for this one...

dataikuapi/dss/ml.py
samronsin: Actually, if the user can do that for `range`, we will want to allow a similar behaviour with `values`. So we need to be careful not to share just pointers and copy the inputs and document the behaviour.

dataikuapi/dss/ml.py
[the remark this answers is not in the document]
mhham: Maybe something like `RangeValues` (since we define the number of values in the range) ?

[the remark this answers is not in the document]
samronsin: `RangeValues` might be a bit confusing since there is a `values` property...

dataikuapi/dss/ml.py
mhham: What about adding a `copy` (or `to_plain`)  method to `NumericalHyperparameterSettings.Range` that outputs a .... `Range` 😅 , so that we can copy from one algo to the other using the same syntactic sugar ?

This would not work for explicit values though, unless we create a `NumericalHyperparameterSettings.Values` or something...

Copying parameters between one algo to another using syntactic sugar requires a bit more thought to be consistent all around, IMO

dataikuapi/dss/ml.py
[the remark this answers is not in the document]
mhham: Some ideas:
* `PlainRange`
* `RangeDefinition`
* Rename `NumericalHyperparameterSettings.Range` to `_Range`

dataikuapi/dss/ml.py
instanceofme: I guess it means arbitration between
- either holding a pointer to a state and being able to do `algo1.hp1.range.min = 3` that edits the parent ~dict
- or copying `.range` to another hp/algo/mltask?

Or maybe when calling the setter with a Range value, changing the pointer?

mhham: We could do both, by adding a specific syntactic sugar for `NumericalHyperparameterSettings.Range` and copying the `min`, `max` and `nb_values` into `set_range`'s arguments, as Nicolas suggested (i.e drop the pointer). I'm not sure whether this is less confusing than adding a kind of `copy` method to `NumericalHyperparameterSettings.Range` as I mentionned above.

For copying `values` between algorithms, after thinking about it again, it seems that it would work out of the box, because calling a `NumericalHyperparameterSettings.values` property returns a `list`

dataikuapi/dss/ml.py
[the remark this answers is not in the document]
samronsin: Thanks ! What do you dislike about `RangeSettings` though ?

dataikuapi/dss/ml.py
mhham: That was before nesting the range inside `NumericalHyperparameterSetting`, resolving it

dataikuapi/dss/ml.py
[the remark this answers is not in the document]
mhham: Nothing wrong with `RangeSettings`, but then use it for the nested class right ?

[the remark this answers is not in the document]
samronsin: Yes !

dataikuapi/dss/ml.py
```diff
@@ -1047,7 +1057,10 @@ def __setattr__(self, attr_name, value):
                 elif isinstance(target, CategoricalHyperparameterSettings):
                     target.set_values(value)
                 elif isinstance(target, NumericalHyperparameterSettings):
-                    raise Exception("Invalid assignment of a NumericalHyperparameterSettings object")
+                    if isinstance(value, list):
+                        target.set_explicit_values(values=value)
+                    elif isinstance(value, Range):
+                        target.set_range(min=value.min, max=value.max, nb_values=value.nb_values)
```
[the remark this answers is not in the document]
samronsin: I agree for `NumericalHyperparameterSettings` but for a categorical hyperparameter, `target.set_values(value)` will be executed which in turn will execute `assert isinstance(value, list)`.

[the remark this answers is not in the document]
samronsin: d7629f3 addresses the issue for `NumericalHyperparameterSettings`, and I don't think we should do anything special for `CategoricalHyperparameterSettings`, or am I missing something ?

dataikuapi/dss/ml.py
samronsin: To simplify things a bit, what about being able to copy the _entire hyperparameter setting_ (except `scaling` and `limits` of course) when doing `algo1.hyperparameter1 = algo2.hyperparameter2` and not allowing any other form of copy ?

mhham: IMO if we manage to create a consistent copy mechanism for the hyperparameter setting fields, then copying the entire setting would be trivial (looping through the fields and copying them one by one). But that would be another syntactic sugar that checks if `value` has the same type as `target` and performs the copy.

For the syntactic sugar that only copies the `NumericalHyperparameterSettings.Range`, I like Nicolas solution that drops the pointer and only copies the `min`, `max` and `nb_values` into `set_range`'s arguments

samronsin: I don't see a situation where the user would only want to copy either a `.range` or a `.values` field from one hyperparameter and not the whole thing. Because they will want to set it either as a `RANGE` or as `EXPLICIT` in which case they only care about either `.range` or `.values`. 

So allowing to copy either the whole hyperparameter setting (`algo1.hyperparameter1 = algo2.hyperparameter2`) or nothing, i.e. forbidding `algo1.hyperparameter1 = algo2.hyperparameter2.range` and `algo1.hyperparameter1 = algo2.hyperparameter2.values` sounds both clearer in terms of API (notice the type consistency at assignment) and simpler in terms of implementation.

dataikuapi/dss/ml.py
```diff
@@ -1064,12 +1064,19 @@ def __setattr__(self, attr_name, value):
                     target.set_values(value)
                 elif isinstance(target, NumericalHyperparameterSettings):
                     if isinstance(value, list):
-                        target.set_explicit_values(values=value)
+                        # algo.hyperparam = [x, y, z]
+                        target.set_explicit_values(values=1)
```
mhham, line 1068: Why `=1` ?

```suggestion
                        target.set_explicit_values(values=value)
```

mhham (changes requested):

dataikuapi/dss/ml.py
samronsin: Weird typo, thanks for catching it !

mhham (approved):

instanceofme (commented):
Basics looks good.

- Maybe we'd implement some tests before merging?
- Could comment/docstring to explain differences between ranges
- Should document / update example notebook
- We could release this for 9.0.X, so
   - Should rebase
   - Must wait for 9.0.1 release before merging

dataikuapi/dss/ml.py
```diff
@@ -812,6 +812,10 @@ def _set_range(self, min=None, max=None, nb_values=None):
                 self._algo_settings[self.name]["range"]["nbValues"] = nb_values
 
     class RangeSettings(object):
+        """
+        Class designed to represent the range of a numerical hyperparameter by pointing to the algorithm settings
```
instanceofme, line 816:
```suggestion
        [Internal] Range of a numerical hyperparameter (points to the algorithm settings)
```

dataikuapi/dss/ml.py
```diff
@@ -890,6 +894,10 @@ def range(self):
 
 
 class Range(object):
+    """
+    Class designed to represent the range of a numerical hyperparameter by storing its attributes (min, max, nb_values)
+    End users should use this class to define explicitly the parameters of the range of a numerical hyperparameter
```
instanceofme, line 899:
```suggestion
    Range of a numerical hyperparameter (min, max, nb_values)
    Use this class to define explicitly the parameters of the range of a numerical hyperparameter
```

instanceofme (commented):
Simplification proposal

samronsin:
Updated the notebook in [CH58941](https://app.clubhouse.io/dataiku/story/58941/public-api-python-client-add-syntactic-sugar-for-numericalhyperparametersettings).

nicolasservel (approved):

PR #120, opened 2021-01-29
Add cross-validation seed 

samronsin:
[CH59982](https://app.clubhouse.io/dataiku/story/59982/public-api-python-client-add-cross-validation-seed-to-hyperparametersearchsettings)

### DIFF
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -473,11 +473,8 @@ class HyperparameterSearchSettings(object):
 
     def _set_cv_seed(self, seed):
         if seed is not None:
-            if not isinstance(seed, int):
-                warnings.warn("HyperparameterSearchSettings ignoring invalid input: seed")
-                warnings.warn("seed must be an integer")
-            else:
-                self._raw_settings["cvSeed"] = seed
+            assert isinstance(seed, int), "HyperparameterSearchSettings invalid input: cvSeed must be an integer"
+            self._raw_settings["cvSeed"] = seed
 
     @property
     def strategy(self):
@@ -566,8 +563,7 @@ class HyperparameterSearchSettings(object):
         :param seed: cross-validation seed for splitting the data during hyperparameter search
         :type seed: int
         """
-        assert isinstance(seed, int)
-        self._raw_settings["cvSeed"] = seed
+        self._set_cv_seed(seed)
 
     def set_kfold_validation(self, n_folds=5, stratified=True, cv_seed=0):
         """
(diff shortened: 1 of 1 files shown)


dataikuapi/dss/ml.py
```diff
@@ -471,6 +471,14 @@ def _set_seed(self, seed):
             else:
                 self._raw_settings["seed"] = seed
 
+    def _set_cv_seed(self, seed):
+        if seed is not None:
+            if not isinstance(seed, int):
+                warnings.warn("HyperparameterSearchSettings ignoring invalid input: seed")
+                warnings.warn("seed must be an integer")
```
mhham, line 478: Why don't we raise a `TypeError` here ? 
This would allow us to use `_set_cv_seed` in `cv_seed.setter` too, instead of manually checking the type there. And  would provide an additional safety in `set_kfold_validation` and `set_single_split_validation`

dataikuapi/dss/ml.py
```diff
@@ -570,8 +595,10 @@ def set_kfold_validation(self, n_folds=5, stratified=True):
                 warnings.warn("stratified must be a boolean")
             else:
                 self._raw_settings["stratified"] = stratified
+        if cv_seed is not None:
```
mhham, line 598: This check seems redundant (already done in `_set_cv_seed`)

dataikuapi/dss/ml.py
```diff
@@ -597,6 +624,8 @@ def set_single_split_validation(self, split_ratio=0.8, stratified=True):
...
```
mhham, line 627: This check seems redundant (already done in `_set_cv_seed`)

dataikuapi/dss/ml.py
```diff
@@ -544,7 +552,24 @@ def validation_mode(self, mode):
...
+    @cv_seed.setter
+    def cv_seed(self, seed):
+        """
+        :param seed: cross-validation seed for splitting the data during hyperparameter search
+        :type seed: int
+        """
+        assert isinstance(seed, int)
+        self._raw_settings["cvSeed"] = seed
+
+    def set_kfold_validation(self, n_folds=5, stratified=True, cv_seed=0):
```
mhham, line 572: Where does the default value `0` come from ?

samronsin: It's the default `int` value.

samronsin: It's arbitrary.

mhham (approved):

PR #121, opened 2021-02-05
Python wrapper for Jupyter Notebooks Public API

william-lukusa:
Story details : [[ch56395]](https://app.clubhouse.io/dataiku/story/56395)

### DIFF
--- a/dataikuapi/dss/notebook.py
+++ b/dataikuapi/dss/notebook.py
@@ -14,7 +14,7 @@ class DSSNotebook(object):
 
     def unload(self, session_id=None):
         """
-        Stop the notebook and release its resources
+        Stop this Jupyter notebook and release its resources
         """
         state = self.get_state()
         if state is None:
@@ -33,12 +33,12 @@ class DSSNotebook(object):
 
     def get_state(self):
         """
-        Get the status of the notebook
+        Get the status of this Jupyter notebook
         """
-        notebook_list = self.client._perform_json("GET",
+        notebook_states = self.client._perform_json("GET",
                                                  "/projects/%s/jupyter-notebooks/" % self.project_key,
                                                  params={"active": False})
-        for notebook in notebook_list:
+        for notebook in notebook_states:
             if notebook.get("name") == self.notebook_name:
                     self.state = notebook
                     return self.state
@@ -46,7 +46,7 @@ class DSSNotebook(object):
 
     def get_sessions(self):
         """
-        Get the list of the running sessions of this notebook
+        Get the list of the running sessions of this Jupyter notebook
         """
         state = self.get_state()
         if state is None:
@@ -57,7 +57,7 @@ class DSSNotebook(object):
 
     def get_content(self):
         """
-        Get the content of this notebook (metadata, cells, nbformat)
+        Get the content of this Jupyter notebook (metadata, cells, nbformat)
         """
         if self.content is None:
             self.content = self.client._perform_json("GET",
@@ -66,7 +66,7 @@ class DSSNotebook(object):
 
     def save(self):
         """
-        Save the content of this notebook
+        Save the content of this Jupyter notebook
         """
         return self.client._perform_json("PUT",
                                          "/projects/%s/jupyter-notebooks/%s" % (self.project_key, self.notebook_name),
@@ -74,7 +74,7 @@ class DSSNotebook(object):
 
     def delete(self):
         """
-        Delete this jupyter notebook and stop all of its active sessions.
+        Delete this Jupyter notebook and stop all of its active sessions.
         """
         return self.client._perform_json("DELETE",
                                          "/projects/%s/jupyter-notebooks/%s" % (self.project_key, self.notebook_name))
--- a/dataikuapi/dss/notebook.py
+++ b/dataikuapi/dss/notebook.py
@@ -41,6 +41,7 @@ class DSSNotebook(object):
         for notebook in notebook_list:
             if notebook.get("name") == self.notebook_name:
                     self.state = notebook
+                    return self.state
         return self.state
 
     def get_sessions(self):
--- a/dataikuapi/dss/notebook.py
+++ b/dataikuapi/dss/notebook.py
@@ -16,44 +16,46 @@ class DSSNotebook(object):
         """
         Stop this Jupyter notebook and release its resources
         """
-        state = self.get_state()
-        if state is None:
+        sessions = self.get_sessions()
+        if sessions is None:
             raise Exception("Notebook isn't running")
-        if state.get('activeSessions', None) is None:
-            raise Exception("Notebook isn't running")
-        if len(state['activeSessions']) == 0:
+        if len(sessions) == 0:
             raise Exception("Notebook isn't running")
         if session_id is None:
-            if len(state['activeSessions']) > 1:
+            if len(sessions) > 1:
                 raise Exception("Several sessions of the notebook are running, choose one")
             else:
-                session_id = state['activeSessions'][0].get('sessionId', None)
+                session_id = sessions[0].get('sessionId', None)
         return self.client._perform_json("DELETE",
                                          "/projects/%s/jupyter-notebooks/%s/sessions/%s" % (self.project_key, self.notebook_name, session_id))
 
-    def get_state(self):
+    def get_state(self, refresh=False):
         """
         Get the status of this Jupyter notebook
+
+        :param bool refresh: if True, get the status of the notebook from the backend
         """
         notebook_states = self.client._perform_json("GET",
-                                                 "/projects/%s/jupyter-notebooks/" % self.project_key,
-                                                 params={"active": False})
-        for notebook in notebook_states:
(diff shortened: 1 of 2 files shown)


dataikuapi/dss/notebook.py
```diff
@@ -27,14 +28,19 @@ def unload(self, session_id=None):
...
         Get the status of the notebook
         """
-        if self.state is None:
-            self.state = self.client._perform_json("GET", "/projects/%s/notebooks/" % self.project_key, params={'notebookName' : self.notebook_name})
+        notebook_list = self.client._perform_json("GET",
+                                                 "/projects/%s/jupyter-notebooks/" % self.project_key,
+                                                 params={"active": False})
+        for notebook in notebook_list:
+            if notebook.get("name") == self.notebook_name:
+                    self.state = notebook
```
adescamps, line 43: You can assign and return directly there instead of finishing the loop.
Is it expected to not cache the state anymore and always retrieve it?

dataikuapi/dss/project.py
```diff
@@ -825,6 +825,54 @@ def new_job_definition_builder(self, job_type='NON_RECURSIVE_FORCED_BUILD'):
...
+        :param bool active: if True, only return currently running jupyter notebooks.
+
+
+        :returns: The list of the notebooks - see as_objects for more information
+        :rtype: list
+        """
+        notebooks = self.client._perform_json("GET", "/projects/%s/jupyter-notebooks/" % self.project_key,
+                                         params={"active": active})
+        if as_objects:
+            return [DSSNotebook(self.client, notebook['projectKey'], notebook['name'], content=notebook) for notebook in notebooks]
```
adescamps, line 847: the call `jupyter-notebooks/` returns a `List<JupyterNotebookListItem>` whereas the DSSNotebook content should be a `NotebookSafeForWritingNew` especially that the DSSNotebook.save() send the content as the body and if you check the call PUT /{notebookName} we expect a NotebookSafeForWritingNew

william-lukusa: Ah yes actually it should be the state, I changed it in the last commit but it was correct before in fact

dataikuapi/dss/notebook.py
william-lukusa: Indeed for the return
For the caching, it's because the state might not have up to date data for `unload`/`get_sessions`. 
For example if we do a `get_notebook` then start a session and try to unload, it will say that the notebook is not running because the sessions in `state` are cached from when there was no sessions.. 
So I don't know what's the best solution :
* Keep the cache but the user must `get_notebook` and `get_state` anyway to have the correct sessions ?
* Keep the changes and always retrieve the latest state ? 
* Or add a `refresh` argument to force refresh the state when we unload/get_sessions ?

adescamps: 👍  I'm just wondering about the fact that we retrieve all the notebooks just to have an updated state (We may miss a endpoint to retrieve the state of just one notebook)

dataikuapi/dss/project.py
```diff
@@ -825,6 +825,54 @@ def new_job_definition_builder(self, job_type='NON_RECURSIVE_FORCED_BUILD'):
...
+    def create_jupyter_notebook(self, notebook_name, notebook_content):
+        """
+        Create a new jupyter notebook.
+
+        :param str notebook_name: the name of the notebook to create
+        :param dict notebook_content: the data of the notebook to create, as a dict.
+            The data will be converted to a JSON string internally.
+            Use ``get_state()`` on a similar existing ``DSSNotebook`` object in order to get a sample definition object
+        """
+        return self.client._perform_json("POST",
```
adescamps, line 872: You should return a handle to the newly create notebook and not the message.
I was playing with it and did:
`notebook = project.create_jupyter_notebook("MyName", { SUPER COMPLICATED DICT 😄  })`
Check for example:
`project.create_recipe()`

Moreover, the Public api returns directly a String /!\ It's a bad practice (I know we all doing that 😄 ) because in case you want to return more info, you will need to break the public API. Return an object that encapsulate the name of the notebook.

dataikuapi/dss/notebook.py
apichery: Yes, I was wondering the same when designing the public API.
I guess we could add an endpoint `GET /project/<projectkey>/notebook/notebookname/sessions` to retrieve the session list

dataikuapi/dss/project.py
apichery: The public API returns
```
{
    "msg": "Created Jupyter notebook 'nb4'"
}
```

adescamps: My bad 😄  I didn't pay too much attention to the WriteMessage method 🙈 . I still think that we should return a DSSNotebook from the method and not directly the message from the API. (We should pay attention to the save because the content will not be loaded)

dataikuapi/dss/notebook.py
```diff
@@ -27,14 +28,20 @@ def unload(self, session_id=None):
...
+        return self.client._perform_json("DELETE",
+                                         "/projects/%s/jupyter-notebooks/%s/sessions/%s" % (self.project_key, self.notebook_name, session_id))
 
     def get_state(self):
         """
         Get the status of the notebook
         """
-        if self.state is None:
-            self.state = self.client._perform_json("GET", "/projects/%s/notebooks/" % self.project_key, params={'notebookName' : self.notebook_name})
+        notebook_list = self.client._perform_json("GET",
```
pbailly, line 38: Should we keep the cache mechanism?
(If you put back the cache, maybe you should add with an optional parameter `clear_cache = false` for when the user know the cache is wrong. Or a `clear_cache()` method)

pbailly, line 38: This variable contains the metadata of the notebook, not a notebook.
The java object is names `JupyterNotebookListItem` which is super confusing so maybe let's not use that.
Maybe something like:
```suggestion
        notebook_states = self.client._perform_json("GET",
```

dataikuapi/dss/notebook.py
```diff
@@ -27,14 +28,20 @@ def unload(self, session_id=None):
                 raise Exception("Several sessions of the notebook are running, choose one")
             else:
                 session_id = state['activeSessions'][0].get('sessionId', None)
-        return self.client._perform_json("DELETE", "/projects/%s/notebooks/" % self.project_key, params={'notebookName' : self.notebook_name, 'sessionId' : session_id})
+        return self.client._perform_json("DELETE",
+                                         "/projects/%s/jupyter-notebooks/%s/sessions/%s" % (self.project_key, self.notebook_name, session_id))
 
     def get_state(self):
         """
         Get the status of the notebook
```
pbailly, line 36: This is (and was) not compatible with SQL notebooks
```suggestion
        Get the status of the Jupyter notebook
```

dataikuapi/dss/project.py
pbailly: I feel weird about the message, after a creation, you should get a reference of the newly created object.
So, to be REST compliant, you should get something like:
```
{
    "id": "nb4",
    "msg": "Created Jupyter notebook 'nb4'"
}
```

dataikuapi/dss/notebook.py
william-lukusa: Should we do a PR to add this endpoint ?

dataikuapi/dss/project.py
william-lukusa: Yes I was looking at other api calls like recipe and they return the object on creation. Should we change it in a PR with the new sessions endpoint @apichery ?

william-lukusa:
⬆️  I created a new PR based on the suggested changes to add an endpoint for sessions and return the object on create/update

dataikuapi/dss/project.py
```diff
@@ -862,16 +862,19 @@ def get_jupyter_notebook(self, notebook_name):
...
+        :returns: A handle to interact with the newly created jupyter notebook
+        :rtype: :class:`~dataikuapi.dss.notebook.DSSNotebook` jupyter notebook handle
         """
-        return self.client._perform_json("POST",
-                                         "/projects/%s/jupyter-notebooks/%s" % (self.project_key, notebook_name),
-                                         body=notebook_content)
+        self.client._perform_json("POST",
+                                  "/projects/%s/jupyter-notebooks/%s" % (self.project_key, notebook_name),
+                                  body=notebook_content)
+        return DSSNotebook(self.client, self.project_key, notebook_name, content=notebook_content)
```
adescamps, line 877: I would have take the content returned by the call and not the one written by the user, because on the backend, we add some new metadata.

dataikuapi/dss/notebook.py
apichery: >Should we do a PR to add this endpoint ?

Yes

dataikuapi/dss/project.py
apichery: >Should we change it in a PR with the new sessions endpoint @apichery ?

Yes

pbailly (approved):
LGTM

adescamps (approved):

PR #123, opened 2021-02-15
update_code_env should not return a DSSFuture

marchelleboid:
Rest API doc also needs to be updated.


cstenac:
Hmm I probably disagree here. It's more the API that should be fixed.

marchelleboid:
> Hmm I probably disagree here. It's more the API that should be fixed.

Ok, I'll make the fix on the public API

PR #124, opened 2021-02-25
[9.0.1] Improve Jupyter Notebook REST API [ch61633]

pbailly:
[[ch61633]](https://app.clubhouse.io/dataiku/story/61633/doubts-about-the-new-jupyter-api)
Also see https://github.com/dataiku/dip/pull/12116

Split the Jupyter Notebook REST API by splitting the metadata (state) part from the main API.


william-lukusa:
I would also add a check in the `save` method to verify the notebook.content is not `None` and throw an error `Notebook content is not initialized/is empty, use get_content() or set the content of the notebook before saving` or something like that 
To have a better error message than the one described in the card

dataikuapi/dss/notebookcontent.py
```diff
@@ -0,0 +1,46 @@
...
+        """
+        return self.get_raw()["cells"]
+
+
+    def save(self):
+        """
+        Save the content of this Jupyter notebook
+        """
+        if self.content is None:
+            raise ValueError("Notebook content is empty, use \"get_content()\" or manually set the content of the notebook before saving")
```
william-lukusa, line 43: > use \"get_content()\"

get_content() doesn't currently set the content, that would be get_raw() but like I said don't know if it's the best place to get the content

dataikuapi/dss/notebook.py
```diff
@@ -9,8 +10,10 @@ def __init__(self, client, project_key, notebook_name, state=None, content=None)
        self.project_key = project_key
        self.notebook_name = notebook_name
        self.state = state
-       self.content = content
-       self.state_is_peek = True
+       if content:
+           self.content = content
+       else:
+           self.content = DSSNotebookContent(self.client, self.project_key, self.notebook_name)
```
william-lukusa, line 16:
```suggestion
      self.content = content
```
Not sure if it makes sense to create a DSSNotebookContent object with no content 
I think maybe `get_content()` should do the api call to get the content and return a DSSNotebookContent object like we do for other DSSObjects, WDYT ?

dataikuapi/dss/notebookcontent.py
```diff
@@ -0,0 +1,46 @@
+from .discussion import DSSObjectDiscussions
+
+class DSSNotebookContent(object):
```
william-lukusa, line 3: Nitpick/Not sure: We could keep it in `notebook.py` like we do for other objects. Don't know what's best

william-lukusa (changes requested):
Some remarks (I'm not sure about) about a few points : 
* setting the content from the get_raw(). 
* having a DSSNotebookContent object but with no content in the init. Should we have the `get_content()` do the api call to get the content and return a DSSNotebookContent object to use ? 
Don't know what you think ? Maybe @apichery has some better opinions on that

pbailly:
`having a DSSNotebookContent object but with no content in the init. Should we have the get_content() do the API call to get the content and return a DSSNotebookContent object to use ?`
If we force the content on the init, this will have the side effect that any DSSNotebook will contain a DSSNotebookContent with its content initialized.
I don't think that is expected, because you may want to simply list your notebook (with the legacy notebook's metadata). In this can you do not want to retrieve your full notebooks.

dataikuapi/dss/notebook.py
pbailly: Since the NotebookContent currently contains only the metadata to retrieve the content, I'm fine with my implementation.
But I don't have a strong feeling about this part.
It may evolve if we decide that the DSSNotebookContent must contains the content.

dataikuapi/dss/notebook.py
```diff
@@ -29,20 +32,15 @@ def unload(self, session_id=None):
         return self.client._perform_json("DELETE",
                                          "/projects/%s/jupyter-notebooks/%s/sessions/%s" % (self.project_key, self.notebook_name, session_id))
 
-    def get_state(self, refresh=False):
+    def get_state(self):
         """
-        Get the status of this Jupyter notebook
-
-        :param bool refresh: if True, get the status of the notebook from the backend
+        Deprecated. Use self.get_content().get_metadata()
```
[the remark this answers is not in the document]
apichery: I would suggest that:
- We kill the `/jupyter-notebooks/<name>/metadata` endpoint in the REST API.
If you want to retrieve the sessions, you call `GET /jupyter-notebooks/<name>/sessions` and if you want to retrieve the metadata, you call `GET /jupyter-notebooks/<name>` and look for the metadata node in the JSON response
- We revert the `get_state()` method to the original code in DSS 8.0 and mark it as deprecated and subject to removal in a future version
```
    def get_state(self):
        """
        Get the status of the notebook
        """
        if self.state is None:
            self.state = self.client._perform_json("GET", "/projects/%s/notebooks/" % self.project_key, params={'notebookName' : self.notebook_name})
        return self.state
```

apichery (changes requested):

dataikuapi/dss/notebook.py
```diff
@@ -1,4 +1,5 @@
 from .discussion import DSSObjectDiscussions
+from .notebookcontent import DSSNotebookContent
 
 class DSSNotebook(object):
```
apichery, line 4: Suggestion.... Actually we may just revert DSSNotebook as it was in DSS 8.0 and declare it as deprecated (as well as all methods returning DSSNotebook instances).

We would create a DSSJupyterNotebook class without the `get_state()` method.

WDYT?

pbailly: Teah, let's do it that way, the legacy stuff is too hard to maintain.

dataikuapi/dss/project.py
```diff
@@ -841,8 +842,10 @@ def list_jupyter_notebooks(self, as_objects=True, active=False):
         :returns: The list of the notebooks - see as_objects for more information
         :rtype: list
         """
-        notebooks = self.client._perform_json("GET", "/projects/%s/jupyter-notebooks/" % self.project_key,
-                                         params={"active": active})
+        notebook_names = self.client._perform_json("GET", "/projects/%s/jupyter-notebooks/" % self.project_key)
+        notebooks = []
+        for notebook_name in notebook_names:
```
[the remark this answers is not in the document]
pbailly: The new version no longer contains metadata.

dataikuapi/dss/notebookcontent.py
pbailly: I will move everything to "jupyterNotebook.py", I don't want to merge legacy and non-legacy code.

dataikuapi/dssclient.py
```diff
@@ -95,14 +95,14 @@ def list_running_notebooks(self, as_objects=True):
         """
         List the currently-running Jupyter notebooks
 
-        :param boolean as_objects: if True, each returned item will be a :class:`dataikuapi.dss.notebook.DSSNotebook`
+        :param boolean as_objects: if True, each returned item will be a :class:`dataikuapi.dss.notebook.DSSJupyterNotebook`
```
apichery, line 98: This part is still a mix between the new and the old. We should have 2 different methods:
- list_running_notebooks() => the old one, returning DSSNotebook objects. Marked as deprecated
- list_jupyter_notebooks(active) => the new one, returning DSSJupyterNotebook objects

apichery (changes requested):

dataikuapi/dss/jupyternootebook.py
```diff
@@ -0,0 +1,101 @@
...
+        return self.content
+
+    def get_cells(self):
+        """
+        Get the cells associated to this Jupyter notebook
+        :rtype: list of cells
+        """
+        return self.content["cells"]
+
+
```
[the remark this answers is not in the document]
pbailly: Ow, I thought you wanted it removed.

dataikuapi/dssclient.py
pbailly: Done. Tell me if you like it.

dataikuapi/dss/jupyternootebook.py
```diff
@@ -0,0 +1,105 @@
...
+        return DSSObjectDiscussions(self.client, self.project_key, "JUPYTER_NOTEBOOK", self.notebook_name)
+
+class DSSNotebookContent(object):
+    """
+    Content of a Jupyter Notebook. Do not create this directly, use :meth:`DSSJupyterNotebook.get_content`
+    """
+
+    """
+    A Python/R/Scala notebook on the DSS instance
+    """
```
william-lukusa, line 70: To be removed

dataikuapi/dss/jupyternootebook.py
```diff
@@ -0,0 +1,105 @@
...
+        return self.content["metadata"]
+
+    def get_cells(self):
+        """
+        Get the cells associated to this Jupyter notebook
+        :rtype: list of cells
+        """
+        return self.content["cells"]
+
+
```
william-lukusa, line 98: Nitpick: format, two blank lines

william-lukusa (commented):
Working great 👍 just two formatting nitpicks

dataikuapi/dss/notebook.py
```diff
@@ -1,94 +1,73 @@
...
     """
-    def __init__(self, client, project_key, notebook_name, state=None, content=None):
-       self.client = client
-       self.project_key = project_key
-       self.notebook_name = notebook_name
-       self.state = state
-       self.content = content
-       self.state_is_peek = True
+    def __init__(self, client, project_key, notebook_name, state=None):
+        warnings.warn("Use DSSJupyterNotebook", DeprecationWarning)
```
apichery, line 10: This warning is useless and annoying. Users do not create DSSNotebook directly.

dataikuapi/dss/project.py
```diff
@@ -856,9 +855,10 @@ def get_jupyter_notebook(self, notebook_name):
         :returns: A handle to interact with this jupyter notebook
         :rtype: :class:`~dataikuapi.dss.notebook.DSSNotebook` jupyter notebook handle
         """
-        notebook_content = self.client._perform_json("GET",
-                                             "/projects/%s/jupyter-notebooks/%s" % (self.project_key, notebook_name))
-        return DSSNotebook(self.client, self.project_key, notebook_name, content=notebook_content)
+        notebook = DSSJupyterNotebook(self.client, self.project_key, notebook_name)
+        # Force content download
+        notebook.get_content()
```
apichery, line 860: Why forcing to download content? Lazy loading is nice ;-)

dataikuapi/dssclient.py
apichery: Looks good now.

apichery (changes requested):
All good except the force load content on `get_jupyter_notebook`

dataikuapi/dss/project.py
pbailly: it was to keep consistency with the previous usage.
I will remove it, then.

apichery (approved):
LGTM

william-lukusa (approved):

dataikuapi/dss/project.py
```diff
@@ -841,12 +841,11 @@ def list_jupyter_notebooks(self, as_objects=True, active=False):
         :returns: The list of the notebooks - see as_objects for more information
         :rtype: list
         """
-        notebooks = self.client._perform_json("GET", "/projects/%s/jupyter-notebooks/" % self.project_key,
-                                         params={"active": active})
+        notebook_names = self.client._perform_json("GET", "/projects/%s/jupyter-notebooks/" % self.project_key, params={"active": active})
```
Tianni-dku, line 844: Your notebook_names is not really a list of string, more `metadata`. An example of printed notebook_name:
[{u'kernelSpec': {u'name': u'python3'}, u'lastModifiedOn': 1616601485243, u'name': u'default', u'language': u'Python', u'creator': u'tian.tian', u'createdOn': 1616601485240, u'tags': [], u'lastModifiedBy': {u'login': u'tian.tian', u'displayName': u'tian.tian'}, u'activeSessions': [{u'notebookName': u'default', u'sessionCreatorDisplayName': u'tian.tian', u'kernelLastActivityTime': 1616598158647, u'sessionId': u'tian.tian__DKU__57eee145-9ee2-41ac-8148-86dc2e175389', u'sessionStartTime': 1616598065590, u'kernelExecutionState': u'idle', u'kernelId': u'3a79aaea-d738-442a-a0eb-db0a8152408f', u'kernelConnections': 1, u'sessionCreator': u'tian.tian', u'projectKey': u'TestPublicAPINotebooks', u'kernelPid': 21350}], u'createdBy': {u'login': u'tian.tian', u'displayName': u'tian.tian'}, u'customFields': {}, u'id': u'default', u'projectKey': u'TestPublicAPINotebooks'}]

dataikuapi/dss/jupyternootebook.py
```diff
@@ -0,0 +1,136 @@
+from .discussion import DSSObjectDiscussions
+
+class DSSJupyterNotebook(object):
+    def __init__(self, client, project_key, notebook_name):
+       self.client = client
```
Tianni-dku, line 5: Also maybe you can change the indentation of line 5, 6, 7

PR #125, opened 2021-03-08
MDG Public API python wrapper

william-lukusa:
Story details : [[ch58204]](https://app.clubhouse.io/dataiku/story/58204)

⚠️ use and merge with https://github.com/dataiku/dip/pull/12168

### DIFF
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -1789,7 +1789,7 @@ class DSSTrainedModelDetails(object):
         files = {'file': fp}
         if self.mltask is not None:
             f = self.mltask.client._perform_json(
-                "POST", "/projects/%s/models/lab/%s/%s/models/%s/model-document/export-documentation-from-custom-template" %
+                "POST", "/projects/%s/models/lab/%s/%s/models/%s/export-documentation-from-custom-template" %
                         (self.mltask.project_key, self.mltask.analysis_id, self.mltask.mltask_id, self.mltask_model_id),
                 files=files)
             return DSSFuture(self.mltask.client, f["jobId"])
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -1811,10 +1811,10 @@ class DSSTrainedModelDetails(object):
         """
         if self.mltask is not None:
             return self.mltask.client._perform_raw(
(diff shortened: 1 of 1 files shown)


dataikuapi/dss/ml.py
```diff
@@ -1752,6 +1752,84 @@ def get_diagnostics(self):
...
+        """
+        Start the export of the model documentation from a docx template (as a file object).
+
+        :param object fp: A file-like object pointing to a template docx file
+        :return: A :class:`~dataikuapi.dss.future.DSSFuture` representing the model document generation process
+        """
+        files = {'file': fp}
+        if self.mltask is not None:
+            f = self.mltask.client._perform_json(
+                "POST", "/projects/%s/models/lab/%s/%s/models/%s/model-document/export-documentation-from-custom-template" %
```
FlorentDev64, line 1792: /model-document/export-documentation-from-custom-template seems not a valid URL

FlorentDev64 (changes requested):

dataikuapi/dss/ml.py
william-lukusa: Ah indeed 👍 changed the url but missed the extra `model-document/`

FlorentDev64 (approved):

FlorentDev64 (approved):
LGTM

FlorentDev64 (commented):
I think that this two routes should be merged into one single route to avoid code duplication.
`"GET", "/projects/%s/models/lab/documentations/%s" % (self.mltask.project_key, export_id))`
`"GET", "/projects/%s/savedmodels/documentations/%s" % (self.saved_model.project_key, export_id))`

Because, we do not need to know where the documentation came from (`savedModels` or 'models/labs`).
Maybe we should have one single route like 
`"GET", "/projects/%sdocumentations/%s" % (self.saved_model.project_key, export_id))`

If this is done, we should move the route to another file.

william-lukusa:
> Because, we do not need to know where the documentation came from (`savedModels` or 'models/labs`). Maybe we should have one single route like `"GET", "/projects/%sdocumentations/%s" % (self.saved_model.project_key, export_id))`

@pbailly @adescamps what do you think ? We discussed it and I still think having the route `/documentations/{exportId}`  in projects might be confusing and keeping it like that is better in the documentation. But I'd like to have some more feedback 😄

PR #130, opened 2021-03-25
Remove unwanted warning in the python doc.

pbailly:
The API was finished, so let's remove these warnings.


william-lukusa (approved):
LGTM 👍

PR #138, opened 2021-03-31
Jupyter notebook / last polish

apichery:


pbailly (approved):
LGTM

dataikuapi/dss/project.py
```diff
@@ -839,11 +839,11 @@ def list_jupyter_notebooks(self, active=False, as_type="object"):
         :returns: The list of the notebooks. If "as_type" is "names", each one as a string, if "as_type" is "objects", each one as a :class:`dataikuapi.dss.notebook.DSSJupyterNotebook`
```
Tianni-dku, line 839: @apichery  Hi the return explanation here `If "as_type" is "names"` is wrong, maybe you forgot to change it. I guess you can just say: 
`If "as_type" is "listitems", then each one as a dict.`

PR #140, opened 2021-04-01
Add publish version method

AgatheG:
[ch61775]

See corresponding PR on dip: https://github.com/dataiku/dip/pull/12374

### DIFF
--- a/dataikuapi/dss/apiservice.py
+++ b/dataikuapi/dss/apiservice.py
@@ -106,7 +106,7 @@ class DSSAPIService(object):
                     f.write(chunk)
                     f.flush()
 
-    def publish_package(self, package_id, published_service_id):
+    def publish_package(self, package_id, published_service_id=None):
         """
         Publish a package on the API Deployer.
 
--- a/dataikuapi/dss/apiservice.py
+++ b/dataikuapi/dss/apiservice.py
@@ -106,3 +106,16 @@ class DSSAPIService(object):
                     f.write(chunk)
                     f.flush()
 
+    def publish_package(self, package_id, published_service_id):
+        """
+        Publish a package on the API Deployer.
+
+        :param string package_id: The identifier of the package
+        :param string published_service_id: The identifier of the API service on the API Deployer where the package will be published.
+            A new published API service will be created if none matches the identifier.
+            If the parameter is not set, the identifier from the current :class:`DSSAPIService` is used.
+        """
+        params = None
+        if published_service_id is not None:
+            params = {"publishedProjectKey": published_service_id}
+        return self.client._perform_json("POST", "/projects/%s/apiservices/%s/packages/%s/publish" % (self.project_key, self.service_id, package_id), params=params)
--- a/dataikuapi/dss/apiservice.py
+++ b/dataikuapi/dss/apiservice.py
@@ -117,5 +117,5 @@ class DSSAPIService(object):
         """
         params = None
         if published_service_id is not None:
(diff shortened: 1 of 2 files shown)


instanceofme (approved):

dataikuapi/dss/apiservice.py
```diff
@@ -106,3 +106,16 @@ def download_package_to_file(self, package_id, path):
...
+
+        :param string package_id: The identifier of the package
+        :param string published_service_id: The identifier of the API service on the API Deployer where the package will be published.
+            A new published API service will be created if none matches the identifier.
+            If the parameter is not set, the identifier from the current :class:`DSSAPIService` is used.
+        """
+        params = None
+        if published_service_id is not None:
+            params = {"publishedProjectKey": published_service_id}
+        return self.client._perform_json("POST", "/projects/%s/apiservices/%s/packages/%s/publish" % (self.project_key, self.service_id, package_id), params=params)
```
marchelleboid, line 121: This publish api service call doesn't return any JSON like the `publish_bundle` version, so you should do `_perform_empty` instead.

dataikuapi/dss/apiservice.py
```diff
@@ -106,3 +106,16 @@ def download_package_to_file(self, package_id, path):
...
+        Publish a package on the API Deployer.
+
+        :param string package_id: The identifier of the package
+        :param string published_service_id: The identifier of the API service on the API Deployer where the package will be published.
+            A new published API service will be created if none matches the identifier.
+            If the parameter is not set, the identifier from the current :class:`DSSAPIService` is used.
+        """
+        params = None
+        if published_service_id is not None:
+            params = {"publishedProjectKey": published_service_id}
```
marchelleboid, line 120: The param should be named `publishedServiceId`

marchelleboid (changes requested):

dataikuapi/dss/apiservice.py
AgatheG: 🙈

dataikuapi/dss/apiservice.py
AgatheG: Oh you're right, will remove the return as well

AgatheG: Although side note, maybe in the long run we'll have this call return something, if we want to have links to the published service in the package list of the API Designer (similar to what we have in the bundle list for now) : https://app.clubhouse.io/dataiku/story/8554/show-which-versions-are-already-pushed-and-replace-by-a-link-that-takes-you-to-the-api-deployer-with-a-preselected-filter

marchelleboid (approved):

AgatheG:
Do we want to merge the refacto from https://github.com/dataiku/dataiku-api-client-python/pull/141?

instanceofme:
Hmm not in this PR. It touches other places than the scope here.

AgatheG:
Yep that was my thought as well. I guess we can merge this PR then

instanceofme:
Maybe we want to wait until `release/9.0` has been published as `9.0.1`, but otherwise LGTM

kevanescence:
@instanceofme @marchelleboid can we merge this ?

marchelleboid:
> @instanceofme @marchelleboid can we merge this ?

Since 9.0.1 has been released, seems like it can be merged now. Might want to double check with @AgatheG though.

kevanescence:
In fact as noticed by @instanceofme  it adds a new method only available in DSS 9.0.2. 
So let's wait for 9.0.2 to land.

PR #146, opened 2021-04-20
project: change get_timeline() item_count param

jfyuen:
Change the parameter to use snake case instead of camel case (more pythonic and in consistent with other parameters).

Part of [ch62224](https://app.clubhouse.io/dataiku/story/62224/api-to-get-project-creation-and-last-modification-date).


FlorentDev64 (approved):
LGTM

PR #148, opened 2021-04-27
ML Task Queues - Train queue + list queues

exil:
This PR adds two calls to the python API:
- `train_queue` - Starts running the queue of the ML task
- `list_mltask_queues` - Lists all the existing ML tasks in a project

[ch63630]

### DIFF
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -3328,15 +3328,15 @@ class DSSMLTask(object):
         self.client._perform_empty(
             "DELETE", "/projects/%s/models/lab/%s/%s/models/%s" % (self.project_key, self.analysis_id, self.mltask_id, model_id))
 
-    def resume_queue(self):
+    def train_queue(self):
         """
-        Resumes a paused queue
+        Trains a queue
 
-        :return: A dict including the sessionID of the resumed queue
+        :return: A dict including the next sessionID to be trained in the queue
         :rtype dict
         """
         return self.client._perform_json(
-            "POST", "/projects/%s/models/lab/%s/%s/actions/resumeQueue" % (self.project_key, self.analysis_id, self.mltask_id))
+            "POST", "/projects/%s/models/lab/%s/%s/actions/trainQueue" % (self.project_key, self.analysis_id, self.mltask_id))
 
     def deploy_to_flow(self, model_id, model_name, train_dataset, test_dataset=None, redo_optimization=True):
         """
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -3328,6 +3328,16 @@ class DSSMLTask(object):
         self.client._perform_empty(
             "DELETE", "/projects/%s/models/lab/%s/%s/models/%s" % (self.project_key, self.analysis_id, self.mltask_id, model_id))
 
+    def resume_queue(self):
+        """
+        Resumes a paused queue
+
+        :return: A dict including the sessionID of the resumed queue
+        :rtype dict
+        """
+        return self.client._perform_json(
+            "POST", "/projects/%s/models/lab/%s/%s/actions/resumeQueue" % (self.project_key, self.analysis_id, self.mltask_id))
+
     def deploy_to_flow(self, model_id, model_name, train_dataset, test_dataset=None, redo_optimization=True):
         """
         Deploys a trained model from this ML Task to a saved model + train recipe in the Flow.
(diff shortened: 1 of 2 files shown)


dataikuapi/dss/project.py
```diff
@@ -591,6 +591,14 @@ def get_ml_task(self, analysis_id, mltask_id):
         """
         return DSSMLTask(self.client, self.project_key, analysis_id, mltask_id)
 
+    def list_mltask_queues(self):
+        """
+        List all ML task queues in this project
+        
+        Returns:
+            the list of the ML task queues, each one as a JSON object
```
instanceofme, line 599:
```suggestion
            the list of the ML task queues, each one as a dict
```

dataikuapi/dss/project.py
```diff
@@ -591,6 +591,14 @@ def get_ml_task(self, analysis_id, mltask_id):
         """
         return DSSMLTask(self.client, self.project_key, analysis_id, mltask_id)
 
+    def list_mltask_queues(self):
+        """
+        List all ML task queues in this project
```
instanceofme, line 596: Which state for the queues? Is there one queue per MLTask even when there is nothing in it and it's not paused?

dataikuapi/dss/project.py
instanceofme, line 599: Seems like it should have a different format (with `:return:` and `:rtype:`)

instanceofme (changes requested):

dataikuapi/dss/project.py
exil: I guess it's the list of ML task queues that were used in some way since the last time DSS was launched

instanceofme: Indeed not easy to explain as a concept.

Does it include non-paused, empty queues?

exil: If the queue was interacted with at any point while DSS was running, it will be list of queues. Empty queues that were run at some point are still in the list.

instanceofme: I guess that it would be less confusing (and befitting the one known use case) to return only the queues that are non empty and/or paused, as well as some additional info. Will comment on the backend API.

dataikuapi/dss/ml.py
```diff
@@ -3423,3 +3423,14 @@ def guess(self, prediction_type=None, reguess_level=None):
             "PUT",
             "/projects/%s/models/lab/%s/%s/guess" % (self.project_key, self.analysis_id, self.mltask_id),
             params = obj)
+
+
+class DSSMLTaskQueues(object):
+    """
+    Object containing a list of MLTask queues
```
instanceofme, line 3430:
```suggestion
    Iterable listing of MLTask queues
```

dataikuapi/dss/project.py
```diff
@@ -593,12 +593,13 @@ def get_ml_task(self, analysis_id, mltask_id):
...
     def list_mltask_queues(self):
         """
-        List all ML task queues in this project
+        List ML task queues in this project 
         
-        Returns:
-            the list of the ML task queues, each one as a JSON object
-        """
-        return self.client._perform_json("GET", "/projects/%s/models/lab/mltask-queues" % self.project_key)
+        :returns: a :class:`DSSMLTaskQueues` object containing a list of queues as dicts
```
instanceofme, line 598:
```suggestion
        :returns: an iterable :class:`DSSMLTaskQueues` listing of MLTask queues (each a dict)
```

dataikuapi/dss/project.py
```diff
@@ -593,12 +593,13 @@ def get_ml_task(self, analysis_id, mltask_id):
 
     def list_mltask_queues(self):
         """
-        List all ML task queues in this project
+        List ML task queues in this project
```
instanceofme, line 596:
```suggestion
        List non-empty ML task queues in this project 
```

dataikuapi/dss/ml.py
```diff
@@ -3423,3 +3423,14 @@ def guess(self, prediction_type=None, reguess_level=None):
...
+    """
+    def __init__(self, queues):
+        self.queues = queues
```
instanceofme, line 3433: Typically we now build these kind of result objects with the whole dict as argument, and make that dict available via `.get_raw()` (with rtype dict). And for example, if we want to expose `.queues`, that could be a `@property` method that returns the internal dict's `["queues"]` item.

instanceofme (changes requested):

instanceofme (approved):
LGTM once we've merged the backend APIs

PR #155, opened 2021-07-19
add call to grab usage summary

FChataigner:
[ch58301]

no helper object, since it wouldn't bring much compared to just accessing fields in the object

### DIFF
--- a/dataikuapi/dssclient.py
+++ b/dataikuapi/dssclient.py
@@ -743,6 +743,19 @@ class DSSClient(object):
             body = custom_params)
 
     ########################################################
+    # Monitoring
+    ########################################################
+
+    def get_global_usage_summary(self):
+        """
+        Summarize the contents of the instance
+
+        :returns: a summary object
+        """
+        return self._perform_json(
+            "GET", "/admin/monitoring/global-usage-summary")
+
+   ########################################################
     # Variables
     ########################################################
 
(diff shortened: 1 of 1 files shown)


dataikuapi/dssclient.py
```diff
@@ -743,6 +743,19 @@ def log_custom_audit(self, custom_type, custom_params=None):
...
 
     ########################################################
+    # Monitoring
+    ########################################################
+
+    def get_global_usage_summary(self):
+        """
+        Summarize the contents of the instance
+
+        :returns: a summary object
```
marchelleboid, line 753: Just to match the doc of other similar methods, I think we usually say it returns a dict rather than an object.

dataikuapi/dssclient.py
```diff
@@ -743,6 +743,19 @@ def log_custom_audit(self, custom_type, custom_params=None):
             body = custom_params)
 
     ########################################################
+    # Monitoring
+    ########################################################
+
+    def get_global_usage_summary(self):
```
qcastel, line 749: From Marc's comment (https://github.com/dataiku/dip/pull/13032#discussion_r672957865), having an option to request the metrics per project would be cool

qcastel (approved):

marchelleboid (approved):

PR #156, opened 2021-07-20
Task/better code env image rebuild

FChataigner:
[ch47120]


thtrunck (approved):
LGTM.
Do we need to do something for the doc? Or will it be auto-regenerated as part of the 9.0.5 release?

dataikuapi/dss/admin.py
```diff
@@ -747,6 +759,52 @@ def update_packages(self, force_rebuild_env=False):
             raise Exception('Env update failed : %s' % (json.dumps(resp.get('messages', {}).get('messages', {}))))
         return resp
 
+    def update_images(self, env_version=None):
+        """
+        Rebuild the docker image of the code env
+        
+        Note: this call requires an API key with admin rights
```
AFrancani, line 766: Isn't it the case for the other calls as well?

AFrancani (approved):
LGTM

PR #159, opened 2021-07-26
Include LightGBM to the list of ML algorithms in the public API

dumontal:
See [ch67573](https://app.clubhouse.io/dataiku/story/67573/support-the-lightgbm-algorithm-in-the-public-api)

While playing with the API I noticed that exposing the LightGBM hyperparameters was not implemented.


samronsin (approved):
Nice, LGTM !

PR #161, opened 2021-08-12
Chore/dss10 clarify get variable apis

AFrancani:
Clarify the `get_*variables` apis:
- rename `dataikuapi.dssclient.DSSClient.get_variables` to `dataikuapi.dssclient.DSSClient.get_global_variables` to clarify its intention
- recreate `dataikuapi.dssclient.DSSClient.get_variables` as a proxy to `dataikuapi.dssclient.DSSClient.get_global_variables` and deprecate it to inform users they should use the new method
- add `dataikuapi.dssclient.DSSClient.get_resolved_variables` mimicing `dataiku.get_custom_variables` to get access to resolved variables without admin permissions

[ch64060]

### DIFF
--- a/dataikuapi/dss/admin.py
+++ b/dataikuapi/dss/admin.py
@@ -956,7 +956,7 @@ class DSSInstanceVariables(dict):
     """
 
     def __init__(self, client, variables):
-        super().__init__()
+        super(dict, self).__init__()
         self.update(variables)
         self.client = client
 
--- a/dataikuapi/dss/admin.py
+++ b/dataikuapi/dss/admin.py
@@ -950,9 +950,9 @@ class DSSClusterStatus(object):
 
 class DSSInstanceVariables(dict):
     """
-    Dict containing the instance variables. The variables can be modified directly in the dict and persisted using its `save` method.
+    Dict containing the instance variables. The variables can be modified directly in the dict and persisted using its :meth:`save` method.
 
-    Do not create this directly, use :meth:`dataikuapi.dssclient.DSSClient.get_global_variables`
+    Do not create this directly, use :meth:`dataikuapi.DSSClient.get_global_variables`
     """
 
     def __init__(self, client, variables):
--- a/dataikuapi/dss/admin.py
+++ b/dataikuapi/dss/admin.py
@@ -946,3 +946,24 @@ class DSSClusterStatus(object):
         Gets the whole status as a raw dictionary.
         """
         return self.status
+
+
+class DSSInstanceVariables(dict):
+    """
+    Dict containing the instance variables. The variables can be modified directly in the dict and persisted using its `save` method.
+
+    Do not create this directly, use :meth:`dataikuapi.dssclient.DSSClient.get_global_variables`
+    """
+
+    def __init__(self, client, variables):
+        super().__init__()
+        self.update(variables)
+        self.client = client
+
+    def save(self):
+        """
+        Save the changes made to the instance variables.
+
+        Note: this call requires an API key with admin rights.
+        """
+        return self.client._perform_empty("PUT", "/admin/variables/", body=self)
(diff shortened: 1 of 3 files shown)


dataikuapi/dss/admin.py
```diff
@@ -946,3 +946,24 @@ def get_raw(self):
...
+
+class DSSInstanceVariables(dict):
+    """
+    Dict containing the instance variables. The variables can be modified directly in the dict and persisted using its `save` method.
+
+    Do not create this directly, use :meth:`dataikuapi.dssclient.DSSClient.get_global_variables`
+    """
+
+    def __init__(self, client, variables):
+        super().__init__()
```
marchelleboid, line 959: This syntax doesn't seem to work in python 2.7. I think we still want to support python 2.7, so you'll probably need to stick with `super(dict, self).__init__()` (I think that still works in python 3)

dataikuapi/dssclient.py
```diff
@@ -760,18 +760,45 @@ def get_global_usage_summary(self, with_per_project=False):
...
         This call requires an API key with admin rights
-        
-        :returns: a Python dictionary of the instance-level variables
+
+        :returns: A :class:`dataikuapi.dss.admin.DSSInstanceVariables` handle
+        """
+        variables = self._perform_json("GET", "/admin/variables/")
+        return DSSInstanceVariables(self, variables)
+
+    def get_resolved_variables(self, project_key=None, typed=False):
```
marchelleboid, line 780: Even though the CH mentions `DSSClient.get_resolved_variables()`, I think this belongs in `DSSProject`, alongside the existing `DSSProject.get_variables()`.

marchelleboid (changes requested):

dataikuapi/dssclient.py
```diff
@@ -782,6 +791,7 @@ def set_variables(self, variables):
         :param dict variables: the new dictionary of all variables of the instance
 
         """
+        warnings.warn("set_variables is deprecated, please use get_global_variables().save()", DeprecationWarning)
         return self._perform_empty(
             "PUT", "/admin/variables/", body=variables)
```
qcastel, line 796: Should we not try to use `get_global_variables().save()` instead?

dataikuapi/dss/project.py
```diff
@@ -961,6 +961,22 @@ def get_variables(self):
         return self.client._perform_json(
             "GET", "/projects/%s/variables/" % self.project_key)
 
+    def get_resolved_variables(self, typed=False):
+        """
+        Get a dictionary of resolved variables of this project.
+
+        :param bool typed: if True, the variable values will be typed in the returned dict, defaults to False
+        :return: a dictionary with instance and project variables merged
```
qcastel, line 969: I guess this line is not needed

qcastel (approved):

marchelleboid (approved):

dataikuapi/dss/admin.py
```diff
@@ -946,3 +946,24 @@ def get_raw(self):
         Gets the whole status as a raw dictionary.
         """
         return self.status
+
+
+class DSSInstanceVariables(dict):
+    """
+    Dict containing the instance variables. The variables can be modified directly in the dict and persisted using its `save` method.
+
+    Do not create this directly, use :meth:`dataikuapi.dssclient.DSSClient.get_global_variables`
```
marchelleboid, line 955: While checking the doc PR, I noticed that this method link needs to be `dataikuapi.DSSClient.get_global_variables`

dataikuapi/dssclient.py
```diff
@@ -760,18 +760,27 @@ def get_global_usage_summary(self, with_per_project=False):
     ########################################################
 
     def get_variables(self):
+        """
+        Deprecated. Use get_global_variables
```
marchelleboid, line 764: Should we add a method link to the new method for the doc (even if it's right next to it)? Same with deprecated comment in `set_variables`

mhoarau (changes requested):

PR #163, opened 2021-08-13
[ch58209] Add personal API keys APIs

qcastel:
PR associated with the change https://github.com/dataiku/dip/pull/13218

### DIFF
--- a/dataikuapi/dss/admin.py
+++ b/dataikuapi/dss/admin.py
@@ -798,6 +798,39 @@ class DSSGlobalApiKey(object):
             "PUT", "/admin/globalAPIKeys/%s" % self.key,
             body = definition)
 
+class DSSPersonalApiKey(object):
+    """
+    A personal API key on the DSS instance
+    """
+    def __init__(self, client, key):
+        self.client = client
+        self.key = key
+
+    ########################################################
+    # Key deletion
+    ########################################################
+
+    def delete(self):
+        """
+        Delete the api key
+        """
+        return self.client._perform_empty(
+            "DELETE", "/personal-api-keys/%s" % self.key)
+
+    ########################################################
+    # Key description
+    ########################################################
+
+    def get_definition(self):
+        """
+        Get the API key's definition
+
+        Returns:
+            the personal API key definition, as a JSON object
+        """
+        return self.client._perform_json(
+            "GET", "/personal-api-keys/%s" % (self.key))
+
 
 class DSSCluster(object):
     """
--- a/dataikuapi/dssclient.py
+++ b/dataikuapi/dssclient.py
@@ -10,7 +10,7 @@ from .dss.projectfolder import DSSProjectFolder
 from .dss.project import DSSProject
 from .dss.app import DSSApp
 from .dss.plugin import DSSPlugin
-from .dss.admin import DSSUser, DSSOwnUser, DSSGroup, DSSConnection, DSSGeneralSettings, DSSCodeEnv, DSSGlobalApiKey, DSSCluster
+from .dss.admin import DSSUser, DSSOwnUser, DSSGroup, DSSConnection, DSSGeneralSettings, DSSCodeEnv, DSSGlobalApiKey, DSSCluster, DSSPersonalApiKey
 from .dss.meaning import DSSMeaning
 from .dss.sqlquery import DSSSQLQuery
 from .dss.discussion import DSSObjectDiscussions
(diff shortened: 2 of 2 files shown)


dataikuapi/dssclient.py
```diff
@@ -629,6 +629,59 @@ def create_global_api_key(self, label=None, description=None, admin=False):
...
+    def get_personal_api_key(self, key):
+        """
+        Get a specific API key from the user doing the request
+
+        :param str key: the secret key of the desired API key
+        :returns: A :class:`dataikuapi.dss.admin.DSSPersonalApiKey` API key handle
+        """
+        return DSSPersonalApiKey(self, key)
+
+    def create_personal_api_key(self, label):
```
AFrancani, line 667: `label` should not be a positional argument given it is not required

dataikuapi/dss/admin.py
```diff
@@ -798,6 +798,41 @@ def set_definition(self, definition):
...
+    ########################################################
+    # Key description
+    ########################################################
+
+    def get_definition(self):
+        """
+        Get the API key's definition
+
+        Returns:
+            the code env definition, as a JSON object
```
AFrancani, line 831:
```suggestion
            the personal API key definition, as a JSON object
```

dataikuapi/dssclient.py
```diff
@@ -629,6 +629,59 @@ def create_global_api_key(self, label=None, description=None, admin=False):
...
+    def delete_global_api_key(self, key):
+        """
+        Delete a specific global API Key
+
+        Note: this call requires an API key with admin rights
+        """
+        resp = self._perform_http(
+            "DELETE", "/admin/globalAPIKeys/" + key)
+        if resp.status_code != 200:
+            raise Exception('API key creation failed : %s' % resp.json())
```
AFrancani, line 641: `dataikuapi.dss.admin.DSSGlobalApiKey` already has a `delete` method. For consistency, this method should be removed.

AFrancani (changes requested):
Don't forget to update https://github.com/dataiku/dss-doc to reflect the news changes

dataikuapi/dssclient.py
qcastel: This is to be retro-compatible with this function deleted line 964

```
    def create_personal_api_key(self, label):
```

(Initially I didn't have the label parameter but did it that way so it doesn't brake existing scripts)

dataikuapi/dss/admin.py
```diff
@@ -798,6 +798,41 @@ def set_definition(self, definition):
             "PUT", "/admin/globalAPIKeys/%s" % self.key,
             body = definition)
 
+class DSSPersonalApiKey(object):
```
qcastel, line 801: @AFrancani I just saw in your PR that you had to do a doc PR for the introduction of a new class.
I bet I have to do the same for this one?

dataikuapi/dssclient.py
AFrancani: It shouldn't be an issue, `label` could be called either by doing `create_personal_api_key(label="my label")` or `create_personal_api_key("my label")`

dataikuapi/dss/admin.py
AFrancani: Yep 👍

dataikuapi/dss/admin.py
```diff
@@ -798,6 +798,41 @@ def set_definition(self, definition):
...
+
+    ########################################################
+    # Key deletion
+    ########################################################
+
+    def delete(self):
+        """
+        Delete the api key
+
+        Note: this call requires an API key with admin rights
```
AFrancani, line 817: Should be removed given users can delete their own keys.

AFrancani (changes requested):

dataikuapi/dss/admin.py
qcastel: yes, definitively: it was a copy-paste artefacts

marchelleboid (approved):

AFrancani (approved):

dataikuapi/dssclient.py
```diff
@@ -629,6 +629,48 @@ def create_global_api_key(self, label=None, description=None, admin=False):
...
+        resp = self._perform_json(
+            "POST", "/personal-api-keys/", body={"label": label})
+        if resp is None:
+            raise Exception('API key creation returned no data')
+        if resp.get('messages', {}).get('error', False):
+            raise Exception('API key creation failed : %s' % (json.dumps(resp.get('messages', {}).get('messages', {}))))
+        if not resp.get('id', False):
+            raise Exception('API key creation returned no key')
+        key = resp.get('key', '')
+        return DSSPersonalApiKey(self, key)
```
mhoarau, line 672: We will force people using the result of this method to change their code. Before it was returning a Dict, where they could do `my_created_key['key']`  (same for the other fields).

Now we need the `get_definition()` to get the same dict as before.

We might inherit a dict to avoid the need for our customers to migrate their code.

mhoarau (changes requested):

PR #164, opened 2021-08-17
Clean deletion process for APIDeployerDeployment

hrjn:
In the API Deployer, deleting a Deployment that is in the `enabled` state (i.e. deployed on an Infrastructure) _without disabling it first_ will remove the Deployment from the UI _but not from the Infrastructure_. This is problematic since users may have to deal with dangling deployments on their Infrastructure(s) not visible on the UI anymore.

This PR introduces a safer way to programmatically delete a Deployment, by first checking if it is enabled, and then:
- disabling the Deployment if the `force` flag is set to `True`,
- throwing an exception prompting the user to disable the Deployment if `force` is set to `False`.

### DIFF
--- a/dataikuapi/dss/apideployer.py
+++ b/dataikuapi/dss/apideployer.py
@@ -339,8 +339,8 @@ class DSSAPIDeployerDeployment(object):
         Deletes this deployment. The disable_first flag automatically disables the deployment
         before its deletion.
 
-        :param boolean disable_first: If True, automatically disables deployment.
-        If False, attempting to delete an enabled deployment will raise an Exception.
+        :param boolean disable_first: If True, automatically disables this deployment before deleting it.
+        If False, will raise an Exception if this deployment is enabled.
 
         """
 
@@ -352,7 +352,7 @@ class DSSAPIDeployerDeployment(object):
             settings = self.get_settings()
             settings.set_enabled(enabled=False)
             settings.save()
-        return self.client._perform_empty(
+        self.client._perform_empty(
                 "DELETE", "/api-deployer/deployments/%s" % (self.deployment_id))
 
                 
@@ -532,7 +532,7 @@ class DSSAPIDeployerService(object):
 
         You may only delete a service if it has no deployments on it anymore.
         """
-        return self.client._perform_empty(
+        self.client._perform_empty(
             "DELETE", "/api-deployer/services/%s" % (self.service_id))
 
 
--- a/dataikuapi/dss/apideployer.py
+++ b/dataikuapi/dss/apideployer.py
@@ -334,15 +334,29 @@ class DSSAPIDeployerDeployment(object):
 
         return DSSFuture(self.client, future_response.get('jobId', None), future_response)
 
-    def delete(self):
+    def delete(self, force=False):
         """
-        Deletes this deployment
+        Deletes this deployment.
+
+        :param boolean force: If True, automatically disables deployment before deletion
 
         You may only delete a deployment if it is disabled and has been updated after disabling it.
         """
-        return self.client._perform_empty(
-            "DELETE", "/api-deployer/deployments/%s" % (self.deployment_id))
 
+        # Check is the deployment is disabled
+        is_enabled = self.get_status().light_status.get("deploymentBasicInfo").get("enabled")
+        if (is_enabled and force) or (not is_enabled):
+            if force:
+                settings = self.get_settings()
+                settings.set_enabled(enabled=False)
+                settings.save()
+            return self.client._perform_empty(
+                "DELETE", "/api-deployer/deployments/%s" % (self.deployment_id))
+        else:
+            raise Exception("Deployment {} deletion failed: deployment must be disabled first.".format(self.deployment_id))
+
+                
+            
 
 class DSSAPIDeployerDeploymentSettings(object):
     """
--- a/dataikuapi/dss/apideployer.py
+++ b/dataikuapi/dss/apideployer.py
@@ -334,26 +334,26 @@ class DSSAPIDeployerDeployment(object):
 
         return DSSFuture(self.client, future_response.get('jobId', None), future_response)
 
-    def delete(self, force=False):
+    def delete(self, disable_first=False):
         """
-        Deletes this deployment.
+        Deletes this deployment. The disable_first flag automatically disables the deployment
+        before its deletion.
 
-        :param boolean force: If True, automatically disables deployment before deletion
+        :param boolean disable_first: If True, automatically disables deployment.
+        If False, attempting to delete an enabled deployment will raise an Exception.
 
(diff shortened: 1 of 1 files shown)


AgatheG:
FYI: we had a card for this issue (https://app.clubhouse.io/dataiku/story/55810/deployer-deleting-a-deployment-does-not-delete-the-deployed-object). Looks like your PR offers a welcome partial/code-only solution 😁

hrjn:
It's indeed more of a "temporary" solution (also, I plead guilty on not checking for existing CH cards before 😅  )

I figured it would be easier to add it directly in the current version of the API client instead of writing a dedicated code sample. Also, I didn't consider the "project deployment deletion" part here.

If it's not that useful/relevant at the moment, I'm happy to kill this 🙂  If it needs some adjustments, I'm happy to improve this 🙂  Just let me know ;)

AgatheG:
Oh no actually I am glad you chose to tackle this 🎉 This is a first step toward implementing it "within DSS". I fully agree with you that having it in the Python API feels cleaner/more robust than providing a sample code. I guess we might want to open another PR to do something similar on the Project deployer side.

I am doing a small review of your code 😀

dataikuapi/dss/apideployer.py
```diff
@@ -334,15 +334,29 @@ def start_update(self):
...
+        Deletes this deployment.
+
+        :param boolean force: If True, automatically disables deployment before deletion
 
         You may only delete a deployment if it is disabled and has been updated after disabling it.
         """
-        return self.client._perform_empty(
-            "DELETE", "/api-deployer/deployments/%s" % (self.deployment_id))
 
+        # Check is the deployment is disabled
```
AgatheG, line 346:
```suggestion
        # Check if the deployment is disabled
```

dataikuapi/dss/apideployer.py
```diff
@@ -334,15 +334,29 @@ def start_update(self):
...
 
-    def delete(self):
+    def delete(self, force=False):
         """
-        Deletes this deployment
+        Deletes this deployment.
+
+        :param boolean force: If True, automatically disables deployment before deletion
 
         You may only delete a deployment if it is disabled and has been updated after disabling it.
```
AgatheG, line 343: I would put this sentence above on line 339 (right after `Deletes this deployment.`)

dataikuapi/dss/apideployer.py
```diff
@@ -334,15 +334,29 @@ def start_update(self):
 
         return DSSFuture(self.client, future_response.get('jobId', None), future_response)
 
-    def delete(self):
+    def delete(self, force=False):
         """
-        Deletes this deployment
+        Deletes this deployment.
+
+        :param boolean force: If True, automatically disables deployment before deletion
```
AgatheG, line 341:
```suggestion
        :param boolean force: If True, automatically disables deployment before deletion. If False, attempting to delete a still enabled deployment will raise an Exception.
```

dataikuapi/dss/apideployer.py
```diff
@@ -334,15 +334,29 @@ def start_update(self):
 
         return DSSFuture(self.client, future_response.get('jobId', None), future_response)
 
-    def delete(self):
+    def delete(self, force=False):
```
AgatheG, line 337: [Nitpick]
Maybe a slightly clearer naming for the param, like `disable_first`?

dataikuapi/dss/apideployer.py
```diff
@@ -334,15 +334,29 @@ def start_update(self):
...
+        is_enabled = self.get_status().light_status.get("deploymentBasicInfo").get("enabled")
+        if (is_enabled and force) or (not is_enabled):
+            if force:
+                settings = self.get_settings()
+                settings.set_enabled(enabled=False)
+                settings.save()
+            return self.client._perform_empty(
+                "DELETE", "/api-deployer/deployments/%s" % (self.deployment_id))
+        else:
+            raise Exception("Deployment {} deletion failed: deployment must be disabled first.".format(self.deployment_id))
```
AgatheG, line 356: In the current state of the code, we always do a call to `save` when `force=True`, even if deployment is disabled. We could ensure to only disable when needed?

```suggestion
        is_enabled = self.get_status().light_status["deploymentBasicInfo"].get("enabled")
        if is_enabled and not force:
            raise Exception("Deployment {} deletion failed: deployment must be disabled first.".format(self.deployment_id))
        if is_enabled:
            settings = self.get_settings()
            settings.set_enabled(enabled=False)
            settings.save()
        self.client._perform_empty(
                "DELETE", "/api-deployer/deployments/%s" % (self.deployment_id))
```

AgatheG (changes requested):
Small comments, overall minor
Thanks again for taking care of this 💪

hrjn:
Done with the small fixes, thx !

dataikuapi/dss/apideployer.py
```diff
@@ -334,15 +334,29 @@ def start_update(self):
...
 
-    def delete(self):
+    def delete(self, disable_first=False):
         """
-        Deletes this deployment
+        Deletes this deployment. The disable_first flag automatically disables the deployment
+        before its deletion.
+
+        :param boolean disable_first: If True, automatically disables deployment.
+        If False, attempting to delete an enabled deployment will raise an Exception.
```
AgatheG, line 343: Sorry for the extra nitpick, just realised that we tend to use `this deployment` in the rest of the doc, so just for consistency (+ a bit of a clearer message for the `False` case, I suggested something a bit too complicated I think)
```suggestion
        :param boolean disable_first: If True, automatically disables this deployment before deleting it.
        If False, will raise an Exception if this deployment is enabled
```

dataikuapi/dss/apideployer.py
```diff
@@ -334,15 +334,29 @@ def start_update(self):
...
+
+        # Check if the deployment is disabled
+        is_enabled = self.get_status().light_status["deploymentBasicInfo"].get("enabled")
+        if is_enabled and not disable_first:
+            raise Exception("Deployment {} deletion failed: deployment must be disabled first.".format(self.deployment_id))
+        if is_enabled:
+            settings = self.get_settings()
+            settings.set_enabled(enabled=False)
+            settings.save()
         return self.client._perform_empty(
```
AgatheG, line 355: Not that this really matters at all but the call does not return anything, so we can remove `return`. But well that's *really* a fly-fucking comment
```suggestion
        self.client._perform_empty(
```

AgatheG (approved):
Two last very, very minor comments (sorry!). Otherwise LGTM 🎉

dataikuapi/dss/apideployer.py
hrjn: Nah, I like polished things, I approve this :)

AgatheG: 😌

PR #165, opened 2021-08-17
Refactoring the MES public API

lpenet:
[ch68817](https://app.clubhouse.io/dataiku/story/68817/refactor-the-public-api-for-model-evaluation-stores?vc_group_by=week&ct_workflow=all&cf_workflow=500000001)

See companion backend PR https://github.com/dataiku/dip/pull/13239 and https://github.com/dataiku/dip/pull/13215 which also trigered some changes.

### DIFF
--- a/dataikuapi/dss/modelevaluationstore.py
+++ b/dataikuapi/dss/modelevaluationstore.py
@@ -122,7 +122,7 @@ class DSSModelEvaluationStore(object):
         :rtype: list of :class:`dataikuapi.dss.modelevaluationstore.DSSModelEvaluation`
         """
         items = self.client._perform_json("GET", "/projects/%s/modelevaluationstores/%s/runs/" % (self.project_key, self.mes_id))
-        return [DSSModelEvaluation(self, item["ref"]["runId"]) for item in items]
+        return [DSSModelEvaluation(self, item["ref"]["runId"], item["created"]) for item in items]
 
     def get_model_evaluation(self, run_id):
         """
@@ -259,13 +259,14 @@ class DSSModelEvaluation:
     Do not create this class directly, instead use :meth:`dataikuapi.dss.DSSModelEvaluationStore.get_model_evaluation`
     """
 
-    def __init__(self, model_evaluation_store, run_id):
+    def __init__(self, model_evaluation_store, run_id, creation_date):
         self.model_evaluation_store = model_evaluation_store
         self.client = model_evaluation_store.client
         # unpack some fields
         self.run_id = run_id
         self.project_key = model_evaluation_store.project_key
         self.mes_id = model_evaluation_store.mes_id
+        self.creation_date = creation_date
 
     def get_full_info(self):
         """
@@ -291,6 +292,14 @@ class DSSModelEvaluation:
         """
         return self.client._perform_json(
             "GET", "/projects/%s/modelevaluationstores/%s/runs/%s/metrics" % (self.project_key, self.mes_id, self.run_id))
+
+    def get_creation_date(self):
+        """
+        Get the creation date, as an epoch
+
+        :return: the creation date, as an epoch
+        """
+        return self.creation_date
 
 
 class DSSModelEvaluationFullInfo:
--- a/dataikuapi/dss/modelevaluationstore.py
+++ b/dataikuapi/dss/modelevaluationstore.py
@@ -110,12 +110,13 @@ class DSSModelEvaluationStore(object):
 
 
     ########################################################
-    # Runs
+    # Model evaluations
     ########################################################
 
     def list_model_evaluations(self, as_type=None):
         """
-        List the model evaluations in this model evaluation store.
+        List the model evaluations in this model evaluation store. The list is sorted
+        by ME creation date.
 
         :returns: The list of the model evaluations
         :rtype: list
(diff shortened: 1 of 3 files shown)


dataikuapi/dss/modelevaluationstore.py
```diff
@@ -110,12 +110,13 @@ def delete(self):
 
 
     ########################################################
-    # Runs
+    # Model evaluations
```
fterrazzoni, line 113: (not part of this PR yet)

After quick check with @instanceofme , we want `project.list_model_evaluation_stores()` to return a list of objects instead of a list of dict by default. There is currently a ` as_type="object"` that does just that by it is not the default. The actual data of the ME should still be accessible using method's of the `DSSModelEvaluationStore`

dataikuapi/dss/modelevaluationstore.py
```diff
@@ -110,12 +110,13 @@ def delete(self):
...
     ########################################################
 
     def list_model_evaluations(self, as_type=None):
         """
-        List the model evaluations in this model evaluation store.
+        List the model evaluations in this model evaluation store. The list is sorted
```
fterrazzoni, line 118: Same comment as above for `project.list_model_evaluation_stores()`: we should return a list of `DSSModelEvaluation`, this makes the API much easier to use & discover.

dataikuapi/dss/modelevaluationstore.py
```diff
@@ -278,80 +286,61 @@ def delete(self):
...
 
     def get_raw(self):
-        return self.settings
+        return self.full_info
 
-    def save(self):
-        return self.client._perform_json(
-                "PUT", "/projects/%s/modelevaluationstores/%s/runs/%s/settings" % (self.project_key, self.mes_id, self.run_id),
-                        body=self.settings)
+    def get_metrics(self):
```
fterrazzoni, line 315: I find this a bit confusing:

```
me.get_full_info().get_metrics()
```
and
```
me.get_metrics()
```

... don't return the same thing and they use a different format

However I don't really see what you can do to clarify that, given how it works under the hood. Maybe one of them you be hidden/"less documented" ? (the 1st is definitely what people will want in practice)

fterrazzoni (changes requested):

dataikuapi/dss/modelevaluationstore.py
fterrazzoni: (I really mean we should get rid of "as_type")

dataikuapi/dss/modelevaluationstore.py
```diff
@@ -148,30 +161,31 @@ def delete_model_evaluations(self, evaluations):
                 obj.append(evaluation['run_id'])
             else:
                 obj.append(evaluation)
-        self.model_evaluation_store.client._perform_json(
+        self.client._perform_json(
                 "DELETE", "/projects/%s/modelevaluationstores/%s/runs/" % (self.project_key, self.mes_id, self.run_id), body=obj)
 
 
-    def create_model_evaluation(self, labels=None, prediction_type=None, model_type=None, model_params=None, data_type=None, data_params=None, metric_params=None, active_classifier_threshold=None):
+    def build(self, job_type="NON_RECURSIVE_FORCED_BUILD", wait=True, no_fail=False):
```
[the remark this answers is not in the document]
lpenet: `no_fail` is a parameter of `start_and_wait()`. When true, an exception is _not_ raised if the job failed.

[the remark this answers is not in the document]
lpenet: Corrected by passing the parameter to `start_and_wait()` + adding docstring for `wait`.

dataikuapi/dss/modelevaluationstore.py
```diff
@@ -278,80 +286,61 @@ def delete(self):
...
+        return self.full_info["evaluation"]["labels"]
 
+    def get_evaluated_model_info(self):
+        """
+        Get info on the evaluated model. Most noticeably, the fullId entry
+        of the dict if the full model id of the evaluated model
+
+        :return: a dict
+        """
+        return self.full_info["evaluation"]["metricParams"]
```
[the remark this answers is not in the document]
lpenet: yes

dataikuapi/dss/modelevaluationstore.py
lpenet: I put it out of consistency with other API. This is the only reason to me: I also see no value in getting dicts instead of objects as building those last ones is not long. So, if @instanceofme agrees to remove `as_type`, I will gladly remove it.

fterrazzoni: He agrees, we discussed this orally

dataikuapi/dss/modelevaluationstore.py
lpenet: Ditto.

dataikuapi/dss/modelevaluationstore.py
lpenet: ok, will update, thanks.

instanceofme: Yes, `as_type` defaulting to dict was the old style, for new APIs we mostly do objects on which one may do `get_raw` to get the underlying dict if they want to.

lpenet: ok, done.

dataikuapi/dss/modelevaluationstore.py
```diff
@@ -110,32 +110,42 @@ def delete(self):
...
+        """
+        Get a handle to interact with the latest model evaluation computed
+
+
+        :returns: A :class:`dataikuapi.dss.modelevaluationstore.DSSModelEvaluation` model evaluation handle
+        """
+
+        latest_run_id = self.client._perform_text(
+            "GET", "/projects/%s/modelevaluationstores/%s/latestRunId" % (self.project_key, self.mes_id))
+        return DSSModelEvaluation(self, latest_run_id)
```
fterrazzoni, line 147: Should not a `DSSModelEvaluation` if the MES is empty (in this case, the `run_id` seems to be "")

lpenet: ok, returns none when there is no latest_run_id

lpenet:
Approved by @fterrazzoni when testing https://github.com/dataiku/dip/pull/13215

PR #168, opened 2021-09-07
[ch58209] Add personal API keys APIs

qcastel:


qcastel:
I noticed that https://github.com/dataiku/dataiku-api-client-python/pull/167 was on the wrong branch name.
Only way to fix this is by creating a new PR. Apologies for making you guys jumping from one PR to another for the same user story! :sweat_smile:

qcastel:
[ch58209] link to clubhouse

dataikuapi/dssclient.py
```diff
@@ -626,8 +634,73 @@ def create_global_api_key(self, label=None, description=None, admin=False):
...
+        """
+        Get a specific API key from the user doing the request
+
+        :param str id: the id the desired API key
+
+        :returns: A :class:`dataikuapi.dss.admin.DSSPersonalApiKey` API key handle
+        """
+        return DSSPersonalApiKey(self, id)
+
+    def create_personal_api_key(self, on_behalf="", label="", description="", as_type='object'):
```
marchelleboid, line 678: There were a comment from @mhoarau on the previous PR about backward compatibility, in this case shouldn't `label` be the second argument, and `as_type` default to `dict`? (sorry if I missed the resolution of that comment)

marchelleboid (changes requested):

dataikuapi/dss/admin.py
```diff
@@ -795,9 +795,53 @@ def set_definition(self, definition):
...
+            the personal API key definition, as a JSON object
+        """
+        return self.client._perform_json(
+            "GET", "/personal-api-keys/%s" % (self.id))
+
+    def update(self, label="", description=""):
+        """
+        Get the API key's definition
+        :param: label string (optional): Label for the API key
+        :param: description string (optional): Description for the API key
```
marchelleboid, line 838: Here you need something like `:param str label: Label for the API key (optional)` and also a empty line between this first param and the `Get the API key's definition` line. Similar for `create_personal_api_key` below.

dataikuapi/dssclient.py
qcastel: I think your comment was meant for the method `create_global_api_key`, as `create_personal_api_key` is actually new.

marchelleboid (approved):

dataikuapi/dss/admin.py
```diff
@@ -752,9 +752,9 @@ class DSSGlobalApiKey(object):
     """
     A global API key on the DSS instance
     """
-    def __init__(self, client, key):
+    def __init__(self, client, id):
```
mhoarau, line 755: We are breaking the existing calls to `dataikuapi.dssclient.DSSClient.get_global_api_key` if we simply do that. 
Cf. this [comment](https://github.com/dataiku/dataiku-api-client-python/pull/167/files#r696773706) on the initial PR (cc @FChataigner ).

And you indeed don't want to call the `/admin/globalAPIKeys` endpoints, we should add another call to resolve the id from the key or keep the previous code with deprecated methods with the key.

dataikuapi/dss/admin.py
```diff
@@ -779,25 +779,67 @@ def get_definition(self):
...
-            definition: the definition for the API key, as a JSON object.                        
+        :param str definition: the definition for the API key, as a JSON object.
+
         """
         return self.client._perform_empty(
-            "PUT", "/admin/globalAPIKeys/%s" % self.key,
+            "PUT", "/admin/global-api-keys/%s" % self.id,
             body = definition)
 
+class DSSPersonalApiKey(object):
```
mhoarau, line 800: Missing one blank line above

dataikuapi/dssclient.py
```diff
@@ -581,16 +581,22 @@ def create_cluster(self, cluster_name, cluster_type='manual', params=None):
...
     ########################################################
 
-    def list_global_api_keys(self):
+    def list_global_api_keys(self, as_type='dict'):
         """
         List all global API keys set up on the DSS instance
 
         Note: this call requires an API key with admin rights
 
         :returns: All global API keys, as a list of dicts
```
mhoarau, line 590: Return message is not exact anymore.

mhoarau (changes requested):

dataikuapi/dss/admin.py
qcastel: good catch, I forgot about this retro-compatibility.
I will add another method.

Not really keen to `resolve the id from the key` without deprecating the method as well, as it will imply we support searching for a key using the secret.
This is not ideal and the implementation would be ugly: search feature in REST API are usually design like `/resource/?attributeA=foo` but that means the key will be as a GET parameter (which is bad security wise for secrets). Therefore I should do a POST but I don't really find that approach very sexy, as this would collision with the POST for create.

Anyway you will see what I suggest in my next commit, probably the best compromise at the end.

AFrancani (changes requested):

FChataigner (approved):
LGTM

AFrancani (approved):
👍

dataikuapi/dssclient.py
```diff
@@ -626,8 +649,73 @@ def create_global_api_key(self, label=None, description=None, admin=False):
...
+        """
+        Get a specific API key from the user doing the request
+
+        :param str id_: the id the desired API key
+
+        :returns: A :class:`dataikuapi.dss.admin.DSSPersonalApiKey` API key
+        """
+        return DSSPersonalApiKey(self, id_)
+
+    def create_personal_api_key(self, label="", description="", as_type='dict', on_behalf=""):
```
mhoarau, line 693: I think that we have many issues with the `on_behalf` parameter.

Currently :
1. Anyone can create a personal key on behalf of anyone and most problematically on behalf of admin.
2. Kiss cool effect: if you ask for a `dict` as return type of the creation command, you are now the happy owner of a personal api key of `admin`.
3. If you don't specify `on_behalf` (it's currently optional and defaulted to an empty string, that the backend interprets as not null which force the `onBehalf` to be empty) you have an unusable key that only an admin clean up. 

All of this comes from the backend code but I prefer to have it here also to double check that it's working in the end.

mhoarau (changes requested):
Careful we have many issues that need to be fixed.

qcastel:
Closing in favor of https://github.com/dataiku/dataiku-api-client-python/pull/226

PR #170, opened 2021-09-10
Added a method to compute data drift on a ME

fterrazzoni:
Relates to https://github.com/dataiku/dip/pull/13469

### DIFF
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -1805,7 +1805,7 @@ class DSSTrainedModelDetails(object):
         return self.details
 
     @property
-    def full_model_like_id(self):
+    def full_id(self):
         return self.details["fullModelId"]
 
     def get_raw_snippet(self):
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -2499,6 +2499,10 @@ class DSSTrainedPredictionModelDetails(DSSTrainedModelDetails):
 
         return roc
 
+    @property
+    def _full_model_like_id(self):
+        return self.details["fullModelId"]
+
     def get_performance_metrics(self):
         """
         Returns all performance metrics for this model.
--- a/dataikuapi/dss/ml.py
+++ b/dataikuapi/dss/ml.py
@@ -1804,6 +1804,10 @@ class DSSTrainedModelDetails(object):
         """
         return self.details
 
+    @property
+    def full_model_like_id(self):
+        return self.details["fullModelId"]
+
     def get_raw_snippet(self):
         """
         Gets the raw dictionary of trained model snippet. 
@@ -2498,10 +2502,6 @@ class DSSTrainedPredictionModelDetails(DSSTrainedModelDetails):
             raise ValueError("This model does not have ROC visualization data")
 
         return roc
-
-    @property
-    def _full_model_like_id(self):
-        return self.details["fullModelId"]
 
     def get_performance_metrics(self):
         """
--- a/dataikuapi/dss/modelevaluationstore.py
+++ b/dataikuapi/dss/modelevaluationstore.py
@@ -289,18 +289,27 @@ class DSSModelEvaluation:
         self.client._perform_json(
                 "DELETE", "/projects/%s/modelevaluationstores/%s/runs/" % (self.project_key, self.mes_id), body=obj)
 
-    def compute_data_drift(self, reference_id=None, data_drift_params=None):
+    @property
+    def _full_model_like_id(self):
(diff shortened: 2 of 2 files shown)


dataikuapi/dss/modelevaluationstore.py
```diff
@@ -288,6 +289,22 @@ def delete(self):
         self.client._perform_json(
                 "DELETE", "/projects/%s/modelevaluationstores/%s/runs/" % (self.project_key, self.mes_id), body=obj)
 
+    def compute_data_drift(self, reference_id=None, data_drift_params=None):
+        """
+        Compute data drift against a reference model or model evaluation. The reference is determined automatically unless specified.
```
lpenet, line 294: It seems to me that computing drift against a lab model is not tested. I would make this limitation clearer there.

lpenet (changes requested):

dataikuapi/dss/modelevaluationstore.py
fterrazzoni: Nothing here specifically mentions "Lab models"? What do you want to change exactly?

(allowing or disallowing this is another debate, to have in the backend PR)

instanceofme (approved):

dataikuapi/dss/modelevaluationstore.py
lpenet: Yes it was none mentioned here. I just wanted to make it clear. It seems to me that we decided to have some positive wording. Something like "Compute data drift against a reference saved model version or a model evaluation"

lpenet:
Playing with it, I felt quite incomfortable with the API, despite the fact that it works.
The user can not easily get the referenceId of a ME or a SMV through the API.
We discussed this with @fterrazzoni and agreed that it would be nice to also allow to pass a `ModelEvaluation` or a `DSSTrainedPredictionModelDetails` as a parameter.
It would also be nice to allow the user to access the fullId easily. But as this is an API change, I guess your opinion is required, @instanceofme

dataikuapi/dss/ml.py
```diff
@@ -2499,6 +2499,10 @@ def get_roc_curve_data(self):
 
         return roc
 
+    @property
+    def _full_model_like_id(self):
+        return self.details["fullModelId"]
+
```
lpenet, line 2505: I would set that in the base class, as it should work for clustering models too.

fterrazzoni: I'm not sure as drift won't work on a clustering model ? 

Or maybe, you wanted this method to be called by end users and in this case, we probably want to make it public ? WDYT ?

lpenet: Obviously, it will not be used for drift. :-)

Yes, I think it could be useful to users and we should make it public.

instanceofme (approved):

dataikuapi/dss/ml.py
fterrazzoni: Done

dataikuapi/dss/modelevaluationstore.py
```diff
@@ -288,6 +289,31 @@ def delete(self):
...
+        return "ME-%s-%s-%s"%(self.project_key, self.mes_id, self.run_id)
+
+    def compute_data_drift(self, reference=None, data_drift_params=None):
+        """
+        Compute data drift against a reference model or model evaluation. The reference is determined automatically unless specified.
+
+        :param reference: saved model version or model evaluation to use as reference (optional, accepts ID or Python object)
+        :type reference: Union[str, DSSModelEvaluation, DSSTrainedPredictionModelDetails]
+        :param data_drift_params: data drift computation settings (optional)
+        :return: data drift analysis results, as a JSON object
```
lpenet, line 303: Sorry, I have a last remark. A DSSFuture is returned, not a JSON object. I would write something like:
```
:returns: a :class:`dataikuapi.dss.future.DSSFuture` tracking the progress of the update. Call 
                   :meth:`~dataikuapi.dss.future.DSSFuture.wait_for_result` on the returned object
                   to wait for completion (or failure) and get the result as a JSON object
```

lpenet (changes requested):
Almost good. One last remark on the doc. Seems important as a user could get perplex, else.

dataikuapi/dss/modelevaluationstore.py
fterrazzoni: Handled differently (auto wait by default, much more practical)

lpenet (approved):
LGTM+tested with Python 2.7 and 3.7. Would like to enrich API doc, but will turn that in an opportunity of learning for a newcomer. :-)

PR #181, opened 2021-10-06
Fleet manager API

duckie:
The initial PR was in master. It should be merged in release/9.0 instead.


duckie:
Target moved to 10.0, I wasnt aware.

PR #188, opened 2021-11-01
Allow to optionally delete data from managed folders and/or job and scenario logs …

vojtech2:
…when deleting a project

[sc-71959](https://app.shortcut.com/dataiku/story/71959/ability-to-delete-job-logs-when-deleting-project)

Also see: dataiku/dip#13848

### DIFF
--- a/dataikuapi/dss/project.py
+++ b/dataikuapi/dss/project.py
@@ -85,9 +85,9 @@ class DSSProject(object):
 
         This call requires an API key with admin rights
 
-        :param bool clear_managed_datasets: Should the data of managed datasets be dropped
-        :param bool clear_output_managed_folders: Should the data from managed folders be dropped
-        :param bool clear_job_and_scenario_logs: Should the job and scenario logs be deleted
+        :param bool clear_managed_datasets: Should the data of managed datasets be cleared
+        :param bool clear_output_managed_folders: Should the data of managed folders used as outputs of recipes be cleared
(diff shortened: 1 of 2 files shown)


pbailly (approved):
LGTM, but don't merge before 10.0.0 release.

dataikuapi/dss/project.py
```diff
@@ -79,17 +79,21 @@ def move_to_folder(self, folder):
     # Project deletion
     ########################################################
 
-    def delete(self, drop_data=False):
+    def delete(self, drop_data=False, drop_managed_folders_output_of_recipe=False, delete_job_logs=False):
```
pbailly, line 82:
```suggestion
    def delete(self, drop_data=False, drop_managed_folders_output_of_recipe=False, delete_job_logs=True):
```

dataikuapi/dss/project.py
```diff
@@ -79,17 +79,25 @@ def move_to_folder(self, folder):
     # Project deletion
     ########################################################
 
-    def delete(self, drop_data=False):
+    def delete(self, clear_managed_datasets=False, clear_output_managed_folders=False, clear_job_and_scenario_logs=True, **kwargs):
```
vojtech2, line 82: I found many instances in integration tests (and beyond) where we call this with `drop_data=True` or just with one positional argument `True`. Should I go ahead and change these calls?

vojtech2: Done.

pbailly (approved):
LGTM

PR #190, opened 2021-11-24
9.0: import for ansible support

duckie:
[ch76255]

### DIFF
--- a/dataikuapi/dss/tools/codegen.py
+++ b/dataikuapi/dss/tools/codegen.py
@@ -1,6 +1,6 @@
 import json, copy, re
-from dataikuapi.dss.recipe import *
-from dataikuapi.dss.dataset import *
+from ..recipe import *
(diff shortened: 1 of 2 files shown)


marchelleboid:
I verified that the python client still works as before, and it does. I'm not sure how to reproduce the original bug, so I haven't tested that.

I did notice another absolute import here: https://github.com/dataiku/dataiku-api-client-python/blob/release/9.0/dataikuapi/dss/admin.py#L215 , does it also need to be changed to a relative import?

duckie:
The import you found should be change but has no impact on the ansible use case, it should be done in a PR to change all imports. Its a wider topic. Providing reproduction code in the shortcut card.

AFrancani (approved):
Code LGTM
Successfully tested

marchelleboid (approved):

PR #192, opened 2021-11-29
Added SER recipe creation

pmmayne:
Companion test PR https://github.com/dataiku/dip/pull/14421

### DIFF
--- a/dataikuapi/dss/recipe.py
+++ b/dataikuapi/dss/recipe.py
@@ -1412,7 +1412,7 @@ class StandaloneEvaluationRecipeCreator(DSSRecipeCreator):
         builder.with_input("scored_dataset_to_evaluate")
         builder.with_output_evaluation_store(evaluation_store_id)
 
-        new_recipe = builder.build()
+        new_recipe = builder.create()
 
         # Save the model parameters in the SER settings
 
@@ -1445,12 +1445,12 @@ class StandaloneEvaluationRecipeCreator(DSSRecipeCreator):
         # Add the newly created json payload to the recipe settings and save the recipe
         # Note that with this method, all the settings that were not explicitly set are instead set to their default value.
 
-        ser_settings = ser.get_settings()
+        ser_settings = new_recipe.get_settings()
 
         ser_settings.set_json_payload(ser_payload)
         ser_settings.save()
 
-        ser.run()
+        new_recipe.run()
 
     Output model evaluation store must exist. It can be created using the following:
 
--- a/dataikuapi/dss/recipe.py
+++ b/dataikuapi/dss/recipe.py
@@ -1423,6 +1423,11 @@ class StandaloneEvaluationRecipeCreator(DSSRecipeCreator):
         ser_payload['features'] = [feature_passengerid, feature_ticket, feature_cabin]
 
         ser_settings = ser.get_settings()
+
+        # Add the newly created json payload to the recipe settings
+        # Note that with this method, all the settings that were not explicitly set are instead set to their default value.
+        # e.g. there is an empty cost matrix setting for this recipe because 'cost matrix' is not defined in the payload
+
         ser_settings.set_json_payload(ser_payload)
         ser_settings.save()
 
--- a/dataikuapi/dss/recipe.py
+++ b/dataikuapi/dss/recipe.py
@@ -1395,7 +1395,7 @@ class StandaloneEvaluationRecipeCreator(DSSRecipeCreator):
 
     .. code-block:: python
 
-        # Create a new standalone evaluation of a dataset
+        # Create a new standalone evaluation of a scored dataset
 
         project = client.get_project("MYPROJECT")
         builder = StandaloneEvaluationRecipeCreator("my_standalone_evaluation_recipe", project)
@@ -1404,7 +1404,7 @@ class StandaloneEvaluationRecipeCreator(DSSRecipeCreator):
 
         new_recipe = builder.build()
 
-        # Save the model parameters in the SER settings and run the SER
+        # Save the model parameters in the SER settings
 
         ser_payload = dict(predictionType="BINARY_CLASSIFICATION",
                    targetVariable="Survived",
@@ -1412,9 +1412,15 @@ class StandaloneEvaluationRecipeCreator(DSSRecipeCreator):
                    isProbaAware=True,
                    dontComputePerformance=False)
 
+        # For a classification model with probabilities, the 'probas' section can be filled with the mapping of the class and the probability column
+        # e.g. for a binary classification model with 2 columns: proba_0 and proba_1
+
         class_0 = dict(key=0, value="proba_0")
         class_1 = dict(key=1, value="proba_1")
         ser_payload['probas'] = [class_0, class_1]
+
+        # Change the 'features' settings for this standalone evaluation
+        # e.g. reject the features that you do not want to use in the evaluation
 
         feature_passengerid = dict(name="Passenger_Id", role="REJECT", type="TEXT")
         feature_ticket = dict(name="Ticket", role="REJECT", type="TEXT")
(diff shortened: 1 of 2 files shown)


dataikuapi/dss/recipe.py
```diff
@@ -1376,6 +1376,37 @@ def with_output_evaluation_store(self, mes_id):
...
+    """
+    Builder for the creation of a new "Standalone Evaluate" recipe, from an
+    input dataset
+
+    .. code-block:: python
+
+        # Create a new standalone evaluation of a dataset
+
+        project = client.get_project("MYPROJECT")
+        builder = EvaluationRecipeCreator("my_standalone_evaluation_recipe", project)
```
michaelbourhis, line 1389: Typo here, I think you meant `StandaloneEvaluationRecipeCreator` 😄

michaelbourhis (changes requested):
LGTM appart from the typo in pydoc

dataikuapi/dss/recipe.py
pmmayne: a classic copypasta mistake

dataikuapi/dss/recipe.py
```diff
@@ -1376,6 +1376,37 @@ def with_output_evaluation_store(self, mes_id):
...
+        builder.with_output_evaluation_store(evaluation_store_id)
+
+        new_recipe = builder.build()
+
+    Output model evaluation store must exist. It can be created using the following:
+
+    .. code-block:: python
+
+        evaluation_store_id = project.create_model_evaluation_store("output_model_evaluation").mes_id
+    """
```
lpenet, line 1400: It would be nice to add to the documentation the payload parameters, such as `isProbaAware`, `dontComputePerformance`, etc.
Would also be great to do the same for the evaluation recipe.
It was not already done for this last, and your were right to do something similar for Standalone Evaluation Recipe, but this would be useful, IMHO.

lpenet (changes requested):
Nice. Let's go a bit further on doc.

lpenet:
An issue I forgot to mention in the review yesterday: when you alter settings, default values for cost matrix are no longer present

image

We should either:
- document this behaviour ;
- set back default values if none is defined.

dataikuapi/dss/recipe.py
pmmayne: I agree

dataikuapi/dss/recipe.py
```diff
@@ -1376,6 +1388,66 @@ def with_output_evaluation_store(self, mes_id):
...
+                   targetVariable="Survived",
+                   predictionVariable="prediction",
+                   isProbaAware=True,
+                   dontComputePerformance=False)
+
+        class_0 = dict(key=0, value="proba_0")
+        class_1 = dict(key=1, value="proba_1")
+        ser_payload['probas'] = [class_0, class_1]
+
+        feature_passengerid = dict(name="Passenger_Id", role="REJECT", type="TEXT")
```
lpenet, line 1419: We overall need more comments in this doc. We should not be exhaustive, but at least indicate here that this is needed only if we want to reject (ignore) some features.

dataikuapi/dss/recipe.py
```diff
@@ -1376,6 +1388,66 @@ def with_output_evaluation_store(self, mes_id):
...
+        feature_ticket = dict(name="Ticket", role="REJECT", type="TEXT")
+        feature_cabin = dict(name="Cabin", role="REJECT", type="TEXT")
+
+        ser_payload['features'] = [feature_passengerid, feature_ticket, feature_cabin]
+
+        ser_settings = ser.get_settings()
+
+        # Add the newly created json payload to the recipe settings
+        # Note that with this method, all the settings that were not explicitly set are instead set to their default value.
+        # e.g. there is an empty cost matrix setting for this recipe because 'cost matrix' is not defined in the payload
```
lpenet, line 1429: So, I would give an example in this doc of setting the costMatrix to the default values.

lpenet (changes requested):

michaelbourhis (approved):

dataikuapi/dss/recipe.py
```diff
@@ -1397,6 +1398,75 @@ def with_output_evaluation_store(self, mes_id):
...
+        feature_cabin = dict(name="Cabin", role="REJECT", type="TEXT")
+
+        ser_payload['features'] = [feature_passengerid, feature_ticket, feature_cabin]
+
+        # To set the cost matrix properly, access the 'metricParams' section of the payload and set the cost matrix weights:
+
+        ser_payload['metricParams'] = dict(costMatrixWeights=dict(tpGain=0.4, fpGain=-1.0, tnGain=0.2, fnGain=-0.5))
+
+        # Add the newly created json payload to the recipe settings and save the recipe
+        # Note that with this method, all the settings that were not explicitly set are instead set to their default value.
```
michaelbourhis, line 1446: I don't have this behaviour for cost matrix: values are just deleted in the SER if I remove `metricParams` key from the `ser_payload`.

dataikuapi/dss/recipe.py
```diff
@@ -1397,6 +1398,75 @@ def with_output_evaluation_store(self, mes_id):
...
+    .. code-block:: python
+
+        # Create a new standalone evaluation of a scored dataset
+
+        project = client.get_project("MYPROJECT")
+        builder = StandaloneEvaluationRecipeCreator("my_standalone_evaluation_recipe", project)
+        builder.with_input("scored_dataset_to_evaluate")
+        builder.with_output_evaluation_store(evaluation_store_id)
+
+        new_recipe = builder.build()
```
michaelbourhis, line 1415: I can't find why, but this creates an invalid recipe. If you just do this step and go check the recipe in the GUI, you will have a JS error.
Either we find a way to properly initialise it, or we clearly state it in the documentation.
I would prefer the first option, but I couldn't figure how to do it.

michaelbourhis (changes requested):
I would also fix in ER recipe documentation, line `1374` of `dataikuapi/dss/recipe.py`: 
`builder = project.new_managed_dataset("output_scored")`
Should be:
`builder = project.new_managed_dataset("output_metrics")`

dataikuapi/dss/recipe.py
lpenet: I am very sorry, but I just tested and I got no error in the GUI. Can you please provide additional details?

michaelbourhis: I've just created a SER and stopped after `builder.build()`. I do not update the settings.

The recipe is created and appears in the flow, but when you open it, it's empty and I have a JS error:
image

pmmayne: I managed to reproduce. I am working on it.

pmmayne: @michaelbourhis I pushed a fix on the backend side of DSS, not on the public api. 

It seemed the backend was trying to force the casting of the output to a dataset-type output (only for the publicapi though, everything was working fine when you worked your way through the GUI to create a SER, as you already know.)

I can only invite you to test it again using the `feature/dss100-sc-74814-add-the-ability-to-create-a-standalone-evaluation` branch for both your DSS and your API client.

See: https://github.com/dataiku/dip/pull/14421

michaelbourhis: That's ok now with that fix, thank you!

dataikuapi/dss/recipe.py
```diff
@@ -1397,6 +1398,75 @@ def with_output_evaluation_store(self, mes_id):
...
+
+        ser_payload['metricParams'] = dict(costMatrixWeights=dict(tpGain=0.4, fpGain=-1.0, tnGain=0.2, fnGain=-0.5))
+
+        # Add the newly created json payload to the recipe settings and save the recipe
+        # Note that with this method, all the settings that were not explicitly set are instead set to their default value.
+
+        ser_settings = new_recipe.get_settings()
+
+        ser_settings.set_json_payload(ser_payload)
+        ser_settings.save()
```
michaelbourhis, line 1451: Sorry, one more thing I've just noticed. When you update the settings this way in the notebook, and open the recipe in the GUI, it is in a dirty state and requires to be saved.

michaelbourhis (changes requested):

dataikuapi/dss/recipe.py
michaelbourhis: After digging on that item, we found the same behaviours with other recipes (evaluation, prepare). But for those, it was due to the labels, and it is another subject.
For the SER, the issue comes from completely replacing the settings dict with a new one that do not have `autoOptimizeThreshold` key. In the GUI, this key is added to the recipe desc, and therefore the recipe is considered changed.

To avoid that, and being consistent with other examples, we suggest to avoid putting the dict replacement in the documentation, and replacing it with an update of the existing dict:

`ser_settings = new_recipe.get_settings()`
`ser_json_payload = ser_settings.get_json_payload()`

`ser_json_payload['predictionType'] = "BINARY_CLASSIFICATION"`
`ser_json_payload['targetVariable'] = "Survived"`
`ser_json_payload['predictionVariable'] = "prediction"`
`ser_json_payload['isProbaAware'] = True`
`ser_json_payload['dontComputePerformance'] = False`

`ser_settings.set_json_payload(ser_json_payload)`
`ser_settings.save()`

michaelbourhis (approved):
LGTM. You can ask @instanceofme or @cstenac for a review.

lpenet:
> LGTM. You can ask @instanceofme or @cstenac for a review.

Nope, we shall ask them to review the base branch after merge.

PR #196, opened 2021-12-08
Added a function to retrieve a model from a comparison

pmmayne:
An integration test is going to be submitted shortly after this PR is merged.

At the moment, the integration test cannot run because of PythonContextSwitcher issues. However, the draft PR can be read at this location:

https://github.com/dataiku/dip/pull/14501

### DIFF
--- a/dataikuapi/dss/project.py
+++ b/dataikuapi/dss/project.py
@@ -894,6 +894,8 @@ class DSSProject(object):
             task_id = full_id.split('-')[3]
             return self.get_ml_task(analysis_id, task_id).get_trained_model_details(full_id)
 
+        raise ValueError("{} is not a valid full model id or full model evaluation id.".format(full_id))
+
     ########################################################
     # Jobs
     ########################################################
--- a/dataikuapi/dss/project.py
+++ b/dataikuapi/dss/project.py
@@ -1,4 +1,5 @@
 import time, warnings, sys, os.path as osp
+import re
 from .dataset import DSSDataset, DSSDatasetListItem, DSSManagedDatasetCreationHelper
 from .modelcomparison import DSSModelComparison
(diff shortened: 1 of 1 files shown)


lpenet (changes requested):
I would be much more conservative there as we are, for instance, in the JS code of the frontend. In the frontend, items are retrieved through our internal API and we can have some expectations on what we test.

Here, we offer an API to a user who can be either a skilled person, or someone randomly using whatever comes at hand.

So, I would use regular expressions and perform full matches. You will find the regex in `FullModelId` and `FullModelEvaluationId`.

I would also raise an exception when no match is found.

I do not think that the `get_comparable` name is adequate. It is meaningful only in the context of MES/MC. Here, we are adding a method to DSSProject I am not the best at naming, but something like `get_from_full_id` reads better to me. Do not hesitate to make a counter-proposal.

pmmayne:
@lpenet you're right and I will add the regex check 

For the renaming, I would suggest `get_model_from_full_id` because what we retrieve is a model (either lab model from analysis, saved model from the flow or a model evaluation from a MES). What's your opinion?

michaelbourhis:
Agreed with @lpenet remarks, and I'm not the best to find name, but I like the one proposed by @pmmayne : we are in the Project context, it's good to add some meaning of what we can retrieve with the method.

pmmayne:
@lpenet I added the pattern matching and changed the name.

However, as I said before, because the API must be usable with Python 2, I cannot use `re.fullmatch` and therefore I decided to add a `\Z` to the regex ensuring that the regex is matching the input String exactly.

In the Integration Test PR, I added tests on the event that a user tries to `get_from_full_id` something that is not the full id of a smv, a lab model or a me.

michaelbourhis (approved):
LGTM

lpenet (approved):
LGTM+Tested

PR #200, opened 2021-12-16
Give MLflowHandle the ability to reload artifact_repository_registry

vthorey:
On `feature/mlflow-experiment-tracking` this code will fail:

```python
from dataikuapi import DSSClient
import mlflow
import tempfile

client = DSSClient("http://localhost:8082", apiKey)
project = client.get_project("DEMOMLFLOW")

with client.setup_mlflow(project.project_key):
	mlflow.log_artifact(tempfile.mkstemp()[1])
```

with:

```shell
~/devmisc/mlflow/mlflow/store/artifact/artifact_repository_registry.py in get_artifact_repository(self, artifact_uri)
     68                 "Could not find a registered artifact repository for: {}. "
     69                 "Currently registered schemes are: {}".format(
---> 70                     artifact_uri, list(self._registry.keys())
     71                 )
     72             )

MlflowException: Could not find a registered artifact repository for: dss-managed-folder:////0/MC8q69UY/artifacts. Currently registered schemes are: ['', 'file', 's3', 'gs', 'wasbs', 'ftp', 'sftp', 'dbfs', 'hdfs', 'viewfs', 'runs', 'models', 'http', 'https']
```

This is because `import mlflow` will register artifacts repositories, including those from plugins, at import. However, we register our `PluginDSSManagedFolderArtifactRepository` in the `setup_client` so after the import of MLflow in this example. 
Hence this PR add the ability for `setup_client` to reload `mlflow.store.artifact.artifact_repository_registry` module. The try excepts handles python2 and 3 (`reload` is a builtin in python2).

The above example should work out of the box on this branch.

Companion DIP branch: https://github.com/dataiku/dip/pull/14636

### DIFF
--- a/dataikuapi/dss_plugin_mlflow/utils.py
+++ b/dataikuapi/dss_plugin_mlflow/utils.py
@@ -38,11 +38,9 @@ class MLflowHandle:
         sys.path.insert(0, self.tempdir)
 
         # Reload the artifact_repository_registry in case MLflow was imported beforehand
-        import mlflow
-        try:  # python 3.4+
+        if sys.version_info > (3, 4):
             from importlib import reload
-        except ImportError:
-            pass
+        import mlflow
         reload(mlflow.store.artifact.artifact_repository_registry)
 
         # Setup authentication
--- a/dataikuapi/dss_plugin_mlflow/utils.py
+++ b/dataikuapi/dss_plugin_mlflow/utils.py
@@ -37,6 +37,14 @@ class MLflowHandle:
             )
         sys.path.insert(0, self.tempdir)
 
+        # Reload the artifact_repository_registry in case MLflow was imported beforehand
+        import mlflow
+        try:  # python 3.4+
+            from importlib import reload
+        except ImportError:
+            pass
+        reload(mlflow.store.artifact.artifact_repository_registry)
+
         # Setup authentication
         if client._session.auth is not None:
             self.mlflow_env.update({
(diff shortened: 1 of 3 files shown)


lpenet:
@jfyuen can you please quickly tell us if the method to handle different python versions is in line with what we usually do in DSS?

lpenet (commented):
Can you please add an integration test?

dataikuapi/dss_plugin_mlflow/utils.py
```diff
@@ -37,6 +37,14 @@ def __init__(self, client, project_key, managed_folder="mlflow_artifacts", host=
             )
         sys.path.insert(0, self.tempdir)
 
+        # Reload the artifact_repository_registry in case MLflow was imported beforehand
+        import mlflow
+        try:  # python 3.4+
+            from importlib import reload
+        except ImportError:
+            pass
+        reload(mlflow.store.artifact.artifact_repository_registry)
```
jfyuen, line 46: Isn't there a "reload" or reset stuff in mlflow?
Anyway, in the dss packages, we prefer using `sys.version_info` instead of checking `ImportError` (although both works)

vthorey:
- [ ] use check on sys.version_info instead of try/except
- [ ] Move setup_mlflow into DSSProject
- [ ] move import of base64 for consistency
- [ ] add integration test in DIP

dataikuapi/dss_plugin_mlflow/utils.py
vthorey: 👍 I have not seen any reload but I am not surprised since normally plugins are installed while installing the package with pip.
I will use sys.versio_info to check on the python version.

PR #201, opened 2021-12-28
Check server side that a MLflow saved model name is not empty when creating or updating

fcandela-dku:
Story details: https://app.shortcut.com/dataiku/story/76185


vthorey (approved):
👍

PR #206, opened 2022-01-06
MLflow for DSS 10.0.3

lpenet:
Public python API changes related to the shortcut cards of the 10.0.3 - MLflow epic https://app.shortcut.com/dataiku/epic/74902?cf_workflow=500000001&vc_group_by=day&ct_workflow=all

Corresponding dip PR https://github.com/dataiku/dip/pull/14742

### DIFF
--- a/dataikuapi/dss/managedfolder.py
+++ b/dataikuapi/dss/managedfolder.py
@@ -2,6 +2,7 @@ from ..utils import DataikuException
 from ..utils import DataikuUTF8CSVReader
 from ..utils import DataikuStreamedHttpUTF8CSVReader
 import json
+import os
 from requests import utils
 from .metrics import ComputedMetrics
 from .future import DSSFuture
@@ -99,7 +100,7 @@ class DSSManagedFolder(object):
     def put_file(self, path, f):
         """
         Upload the file to the managed folder
-        
+
         Args:
             f: the file contents, as a stream
             path: the path of the file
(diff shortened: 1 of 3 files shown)


dataikuapi/dss/managedfolder.py
```diff
@@ -108,6 +109,21 @@ def put_file(self, path, f):
                 "POST", "/projects/%s/managedfolders/%s/contents/%s" % (self.project_key, self.odb_id, utils.quote(path)),
                 "", f)
 
+    def upload_folder(self, path, folder):
+        """
+        Upload folder and its content as path in the managed folder.
```
cstenac, line 114: I think it would be good to document the detail of behavior that it puts "at the location", not "as child of location",

i.e. that upload_folder("target", "source") will result in "target" containing the content of "source", not in "target" containing "source"

dataikuapi/dss/managedfolder.py
```diff
@@ -108,6 +109,21 @@ def put_file(self, path, f):
...
+
+        :param str path: the destination path of the folder in the managed folder
+        :param str folder: path  (absolute or relative) of the source folder to upload
+        """
+        real_root = os.path.realpath(folder)
```
cstenac, line 119: not sure if this is needed. It may actually make it more complex, I think os.walk will already natively return paths relative to the root

vthorey: Indeed, I will simplify it

dataikuapi/dss/managedfolder.py
vthorey: I rephrased into:

> Upload the content of a folder at path in the managed folder.
> 
> Note: upload_folder("target", "source") will result in "target" containing the content
> of "source", not in "target" containing "source".

dataikuapi/dss/savedmodel.py
```diff
@@ -315,9 +345,10 @@ def evaluate(self, dataset_ref):
 
         :meth:`set_core_metadata` must be called before you can evaluate a dataset
 
-        :param str dataset_ref: Name of the evaluation dataset to use (either a dataset name or "PROJECT.datasetName")
+        :param str dataset_ref: Evaluation dataset to use (either a dataset name, "PROJECT.datasetName", DSSDataset instance or dataiku.Dataset instance)
```
cstenac, line 348: ultra minor nitpick: make "DSSDataset" and "dataiku.Dataset" `:class:` so that they are hyperlinked in the doc

cstenac (approved):

dataikuapi/dss/savedmodel.py
lpenet: ok

PR #209, opened 2022-01-19
MLFlowVersionHandler.set_core_metadata implement support for features_list

GregGambino:
In this PR, we add the support for the features_list argument for the set_core_metadata in the MLFLowVersionHandler.

**NB** : This PR is linked to https://github.com/dataiku/dip/pull/14866

### DIFF
--- a/dataikuapi/dss/savedmodel.py
+++ b/dataikuapi/dss/savedmodel.py
@@ -315,6 +315,9 @@ class MLFlowVersionHandler:
         :param list features_list: List of {"name": "feature_name", "type": "feature_type"}
         """
 
+        if features_list is not None and get_features_from_dataset is not None:
+            raise Exception("The information of the features should come either from the features_list or get_features_from_dataset, but not both.")
+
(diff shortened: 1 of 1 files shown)


dataikuapi/dss/savedmodel.py
```diff
@@ -330,9 +333,10 @@ def set_core_metadata(self,
         #if get_features_from_signature:
         #    raise Exception("Get features from signature is not yet implemented")
 
-        # TODO: Add support for features_list, with validation
+        if features_list is not None:
+            metadata["features"] = features_list
```
lpenet, line 337: One common mistake that our beloved users do is passing a list of strings as `features_list`, the strings being the names of the features. And it tooks us some time to spot the error.

So, I would also have a check on the content of the `features_list`, checking that each item in the list is an object with a `name` and a `type` properties.

lpenet (changes requested):
An additional request. :-) This addition request should be the subject of an integration test in the companion DIP PR.

dataikuapi/dss/savedmodel.py
GregGambino: Good call I did it !

vthorey (approved):
LGTM !

lpenet (approved):
LGTM+Tested on catboost

PR #215, opened 2022-02-16
[sc-72323] OAuth2 API node - Allow sending a bearer token

qcastel:
Allowing to send a bearer token, so we can give a python sample code

### DIFF
--- a/dataikuapi/base_client.py
+++ b/dataikuapi/base_client.py
@@ -5,8 +5,9 @@ from requests.auth import HTTPBasicAuth
 from .utils import DataikuException
 
 class DSSBaseClient(object):
-    def __init__(self, base_uri, api_key=None, internal_ticket=None):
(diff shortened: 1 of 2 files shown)


qcastel:
[sc-72323]

dataikuapi/base_client.py
```diff
@@ -5,8 +5,9 @@
 from .utils import DataikuException
 
 class DSSBaseClient(object):
-    def __init__(self, base_uri, api_key=None, internal_ticket=None):
+    def __init__(self, base_uri, api_key=None, bearer_token=None, internal_ticket=None):
```
thtrunck, line 8: Isn't that risky to add a new parameter in the middle of the signature?

AFrancani (changes requested):
I agree with Theo's comment, apart from that LGTM

AFrancani (approved):

thtrunck (approved):

PR #218, opened 2022-02-21
Feature/dss100 sc 79583 drift ser create a ser with a reference dataset

pmmayne:
Companion PR https://github.com/dataiku/dip/pull/15247


michaelbourhis (approved):
LGTM

PR #219, opened 2022-02-21
[sc-80919] new PAPI - dataset get full info

NawreCode:
new public Python API to retrieve all the information about a dataset.

### DIFF
--- a/dataikuapi/dss/dataset.py
+++ b/dataikuapi/dss/dataset.py
@@ -1,3 +1,5 @@
+import datetime
+
 from ..utils import DataikuException
 from ..utils import DataikuUTF8CSVReader
 from ..utils import DataikuStreamedHttpUTF8CSVReader
@@ -522,16 +524,15 @@ class DSSDataset(object):
                 "GET", "/projects/%s/datasets/%s/metrics/history/%s" % (self.project_key, self.dataset_name, 'NP' if len(partition) == 0 else partition),
                 params={'metricLookup' : metric if isinstance(metric, str) or isinstance(metric, unicode) else json.dumps(metric)})
 
-    def get_full_info(self):
+    def get_info(self):
         """
         Retrieve all the information about a dataset
 
-        Returns:
-            a complex JSON object containing all the information about a dataset, such as params, schema, last build infos, status, etc.
+        :returns: a :class:`DSSDatasetInfo` containing all the information about a dataset, such as params, schema, last build infos, status, etc.
+        :rtype: :class:`DSSDatasetInfo`
         """
-        return self.client._perform_json(
-                "GET", "/projects/%s/datasets/%s/info" % (self.project_key, self.dataset_name)
-        )
+        data = self.client._perform_json("GET", "/projects/%s/datasets/%s/info" % (self.project_key, self.dataset_name))
+        return DSSDatasetInfo(self, data)
 
     ########################################################
     # Misc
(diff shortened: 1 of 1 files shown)


dataikuapi/dss/dataset.py
```diff
@@ -523,6 +522,17 @@ def get_metric_history(self, metric, partition=''):
                 "GET", "/projects/%s/datasets/%s/metrics/history/%s" % (self.project_key, self.dataset_name, 'NP' if len(partition) == 0 else partition),
                 params={'metricLookup' : metric if isinstance(metric, str) or isinstance(metric, unicode) else json.dumps(metric)})
 
+    def get_full_info(self):
+        """
+        Retrieve all the information about a dataset
+
+        Returns:
+            a complex JSON object containing all the information on a dataset, such as params, schema, lastbuild infos, status, etc.
```
marchelleboid, line 530:
```suggestion
            a complex JSON object containing all the information about a dataset, such as params, schema, last build infos, status, etc.
```

marchelleboid (approved):

thtrunck (approved):
LGTM

cstenac (changes requested):
* Let's avoid "old-style" APIs where methods directly return raw dicts, as they hard to work with:
    * Create a dataset DSSDatasetInfo
    * get_full_info -> get_info
    * get_info returns a DSSDatasetInfo
    * DSSDatasetInfo has a .raw property returning the dict
    * Add accessor properties on DSSDatasetInfo, and notably a `last_build` which returns either a timestamp or `datetime.datetime`
* Beware, in Python you never return "JSON objects", but "dicts"
* Use modern :arg: :return:, .. comments

NawreCode:
@cstenac I've addressed your comments. Tell me if you see anything else which has to be exposed in the `DSSDatasetInfo` class. I've kept it simple and focused on the last build for now.

thtrunck (approved):

dataikuapi/dss/dataset.py
```diff
@@ -859,3 +870,78 @@ def already_exists(self):
...
+
+
+class DSSDatasetInfo(object):
+    """
+    Info class for a DSS dataset (Read-Only).
+    Do not instantiate this class directly, use :meth:`DSSDataset.get_info`
+    """
+
+    def __init__(self, dataset, info):
+        """
```
cstenac, line 882: Nitpick: we don't normally document constructors that we ask customers not to call

dataikuapi/dss/dataset.py
```diff
@@ -859,3 +870,78 @@ def already_exists(self):
...
+    def get_raw(self):
+        """
+        Get the raw dataset full information as a dict
+
+        :return: the raw dataset full information
+        :rtype: dict
+        """
+        return self.info
+
+    def get_raw_last_build(self):
```
cstenac, line 898: I don't think this method is useful

dataikuapi/dss/dataset.py
```diff
@@ -859,3 +870,78 @@ def already_exists(self):
...
+        """
+        Get the last build information of the dataset, as a raw dict, or None if there is no last build information.
+
+        :return: the last build information
+        :rtype: dict or None
+        """
+        return self.info.get("lastBuild", None)
+
+    def get_last_build_start_time(self, as_date=False):
```
cstenac, line 907: I would put as_date = True by default.

cstenac: Actually, to be consistent with scenario.py, I would:

* Remove the option as_date.
* Always return a date
* Make it a property

check scenario.py 510 and 517

dataikuapi/dss/dataset.py
```diff
@@ -859,3 +870,78 @@ def already_exists(self):
...
+        :rtype: int or :class:`datetime.datetime` or None
+        """
+        last_build_info = self.info.get("lastBuild", dict())
+        timestamp = last_build_info.get("buildEndTime", None)
+
+        if as_date and timestamp:
+            return datetime.datetime.fromtimestamp(timestamp / 1000)
+        return timestamp
+
+    def is_last_build_successful(self):
```
cstenac, line 937: make it a property. Check what we do in scenario.py (462/466/...)

dataikuapi/dss/dataset.py
```diff
@@ -523,6 +524,16 @@ def get_metric_history(self, metric, partition=''):
                 "GET", "/projects/%s/datasets/%s/metrics/history/%s" % (self.project_key, self.dataset_name, 'NP' if len(partition) == 0 else partition),
                 params={'metricLookup' : metric if isinstance(metric, str) or isinstance(metric, unicode) else json.dumps(metric)})
 
+    def get_info(self):
+        """
+        Retrieve all the information about a dataset
+
+        :returns: a :class:`DSSDatasetInfo` containing all the information about a dataset, such as params, schema, last build infos, status, etc.
```
cstenac, line 531: I would probably not mention that it contains "params, schema, status", since we provide no accessors for these.

Beware, "information" is always singular.

cstenac (changes requested):
Minor feedback, but we can still get it a bit better :)

dataikuapi/dss/dataset.py
NawreCode: removed 👍🏻

dataikuapi/dss/dataset.py
NawreCode: done 👍🏻

dataikuapi/dss/dataset.py
NawreCode: done 👍🏻

dataikuapi/dss/dataset.py
NawreCode: removed 👍🏻

dataikuapi/dss/dataset.py
NawreCode: I guess you're right. Removed 👍🏻

PR #220, opened 2022-02-24
Feature store public API

lpenet:
Companion dip PR: https://github.com/dataiku/dip/pull/15317

This PR adds the public python API of the feature store.

It introduces a `DSSFeatureStore` class, than can be instantiated through the api client, whose unique method allows to list feature groups.

It also add a read property and a method to `DSSDatasetSettings` to check if a dataset is a feature group and to promote (or downgrade) a dataset to the quality of feature group.

### DIFF
--- a/dataikuapi/dss/dataset.py
+++ b/dataikuapi/dss/dataset.py
@@ -741,7 +741,6 @@ class DSSDatasetSettings(DSSTaggableObjectSettings):
     def is_feature_group(self):
         """
         Indicates whether the Dataset is defined as a Feature Group, available in the Feature Store.
-        Changes of this property will be applied when calling :meth:`save` and require the "Manage Feature Store" permission.
 
         :rtype: bool
         """
@@ -749,7 +748,7 @@ class DSSDatasetSettings(DSSTaggableObjectSettings):
 
     def set_feature_group(self, status):
         """
-        Indicates whether the Dataset is defined as a Feature Group, available in the Feature Store.
+        (Un)sets the dataset as a Feature Group, available in the Feature Store.
         Changes of this property will be applied when calling :meth:`save` and require the "Manage Feature Store" permission.
 
         :param status: whether the dataset should be defined as a feature group
--- a/dataikuapi/dss/dataset.py
+++ b/dataikuapi/dss/dataset.py
@@ -737,6 +737,26 @@ class DSSDatasetSettings(DSSTaggableObjectSettings):
     def add_raw_schema_column(self, column):
         self.settings["schema"]["columns"].append(column)
 
+    @property
+    def is_feature_group(self):
+        """
+        Indicates whether the Dataset is defined as a Feature Group, available in the Feature Store.
+        Changes of this property will be applied when calling :meth:`save` and require the "Manage Feature Store" permission.
+
+        :rtype: bool
+        """
+        return self.settings["featureGroup"]
+
+    def set_feature_group(self, status):
+        """
+        Indicates whether the Dataset is defined as a Feature Group, available in the Feature Store.
+        Changes of this property will be applied when calling :meth:`save` and require the "Manage Feature Store" permission.
+
+        :param status: whether the dataset should be defined as a feature group
+        :type status: bool
+        """
+        self.settings["featureGroup"] = status
+
     def save(self):
         self.dataset.client._perform_empty(
                 "PUT", "/projects/%s/datasets/%s" % (self.dataset.project_key, self.dataset.dataset_name),
--- a/dataikuapi/dss/feature_store.py
+++ b/dataikuapi/dss/feature_store.py
@@ -12,6 +12,10 @@ class DSSFeatureGroupListItem(object):
         return self.project_key + "." + self.name
 
     def get_as_dataset(self):
+        """
+        :return: a handle on the dataset
+        :rtype: :class:`dataikuapi.dss.dataset.DSSDataset`
+        """
         return DSSDataset(self.client, self.project_key, self.name)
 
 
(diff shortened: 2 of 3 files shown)


dataikuapi/dss/feature_store.py
```diff
@@ -0,0 +1,14 @@
+class DSSFeatureStore(object):
+    def __init__(self, client):
+        """
+        A handle on the Feature Store. Do not call this methid directly, call """
```
michaelbourhis, line 4: Call 911? 🤣 
The method to call is missing, also, there is a typo in `methid`.

michaelbourhis (changes requested):

dataikuapi/dss/dataset.py
```diff
@@ -737,6 +737,26 @@ def add_time_partitioning_dimension(self, dim_name, period="DAY"):
     def add_raw_schema_column(self, column):
         self.settings["schema"]["columns"].append(column)
 
+    @property
+    def is_feature_group(self):
+        """
+        Indicates whether the Dataset is defined as a Feature Group, available in the Feature Store.
+        Changes of this property will be applied when calling :meth:`save` and require the "Manage Feature Store" permission.
```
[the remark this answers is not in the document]
lpenet: Yes.

dataikuapi/dss/feature_store.py
lpenet: Ha ha. Yes, plus got the wording closer to other such cases.

dataikuapi/dss/dataset.py
```diff
@@ -737,6 +737,26 @@ def add_time_partitioning_dimension(self, dim_name, period="DAY"):
...
+
+        :rtype: bool
+        """
+        return self.settings["featureGroup"]
+
+    def set_feature_group(self, status):
+        """
+        Indicates whether the Dataset is defined as a Feature Group, available in the Feature Store.
```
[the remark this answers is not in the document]
lpenet: ok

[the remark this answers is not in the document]
lpenet: I am not sure that "promote" is the adequate wording. Made another proposal.

dataikuapi/dssclient.py
```diff
@@ -1118,6 +1119,14 @@ def get_object_discussions(self, project_key, object_type, object_id):
         """
         return DSSObjectDiscussions(self, project_key, object_type, object_id)
 
+    ########################################################
+    # Feature Store
+    ########################################################
+    def get_feature_store(self):
+        return DSSFeatureStore(self)
```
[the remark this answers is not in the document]
lpenet: yes, good point.

dataikuapi/dss/feature_store.py
```diff
@@ -0,0 +1,14 @@
...
+        self.client = client
+
+    def list_feature_groups(self):
+        """
+        Get a list of names of datasets defined as feature groups on the DSS instance.
```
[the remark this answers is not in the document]
lpenet: Good point!

michaelbourhis (approved):
LGTM!

dataikuapi/dss/dataset.py
```diff
@@ -737,6 +737,25 @@ def add_time_partitioning_dimension(self, dim_name, period="DAY"):
...
+    @property
+    def is_feature_group(self):
+        """
+        Indicates whether the Dataset is defined as a Feature Group, available in the Feature Store.
+
+        :rtype: bool
+        """
+        return self.settings["featureGroup"]
+
+    def set_feature_group(self, status):
```
cstenac, line 749: Rather make it a @property.setter

dataikuapi/dss/feature_store.py
```diff
@@ -0,0 +1,17 @@
+class DSSFeatureStore(object):
+    def __init__(self, client):
+        """
+        A handle on the Feature Store.
+        Do not create this class directly, use :meth:`DSSClient.get_feature_store`
+        """
+        self.client = client
+
+    def list_feature_groups(self):
```
cstenac, line 9: Since this method is cross-project, it cannot return dataset "names". It may possibly return dataset full names (project + dataset), but in that case, the returned value is fairly unusable.

Please make it return a list of DSSFeatureGroupsListItem which have:

* `.id` (returns full project+name)
* `.get_as_dataset()` which returns a DSSDataset from the correct project

dataikuapi/dss/feature_store.py
```diff
@@ -0,0 +1,17 @@
...
+        """
+        Get a list of names of datasets on which the user has right permissions
+        and that are defined as feature groups in the DSS instance
+
+        :return: list of dataset names
+        :rtype: list of str
+        """
+        return self.client._perform_json("GET", "/feature-store/feature-groups/list")
```
cstenac, line 17: The method for "list feature groups" should be `GET /feature-store/feature-groups`, no /list

cstenac (changes requested):

dataikuapi/dss/feature_store.py
lpenet: It returns indeed full ids, and the only use of those ids is to retrieve the associated dataset.
ok, will do.

dataikuapi/dss/dataset.py
lpenet: I initially used a setter, but the idea was to allow specifying other params if the featureGroup has more options... But ok. :-)

lpenet: So, we eventually decided to leave it as is.

dataikuapi/dss/feature_store.py
lpenet: ok

PR #221, opened 2022-03-01
Dont auto create managed folder in setup_mlflow

alarive:
Stop creating managed folder automatically in setup_mlflow.

Tested manually and by running the integration/python/scenario/ml/prediction/mlflow/experiment_tracking/test_autolog.py test.

Related to:
- https://github.com/dataiku/dip/pull/15360
- https://github.com/dataiku/dku-contrib-private/pull/151

### DIFF
--- a/dataikuapi/dss_plugin_mlflow/utils.py
+++ b/dataikuapi/dss_plugin_mlflow/utils.py
@@ -1,5 +1,5 @@
-from ..dss.utils import AnyLoc
 from ..utils import DataikuException
+from ..dss.managedfolder import DSSManagedFolder
 import logging
 import os
 import sys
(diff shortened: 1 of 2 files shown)


alarive:
dip on demand: http://dku13.dataiku.com:9180/job/dip-on-demand/7726/

dataikuapi/dss_plugin_mlflow/utils.py
```diff
@@ -6,7 +9,7 @@
 
 
 class MLflowHandle:
-    def __init__(self, client, project_key, managed_folder_name, host=None):
+    def __init__(self, client, project, managed_folder, host=None):
```
lpenet, line 12: It seems to me that it would be simpler if `managed_folder` could only be a handle on a MF, and not a MF name.
We would avoid checking that it exists, etc. In this case, we would only check that `managed_folder` is an instance of `DSSManagedFolder` and add its type to the docstring.

lpenet (commented):
Not really sure of the way we specify the managed folder. Please see review comment. :-)

alarive:
New dip on demand: http://dku13.dataiku.com:9180/job/dip-on-demand/7754/

lpenet:
DOD is ok.

image

lpenet (approved):
LGTM+DOD ok. Let's check with @cstenac that he likes the `find_managed_folder_by_name` method.

alarive:
DOD still ok: 
image

lpenet (approved):
LGTM

PR #223, opened 2022-03-15
Feature/feature store

lpenet:
Public API of feature store


lpenet:
Already discussed with Clément.

PR #226, opened 2022-03-29
[sc58209] Add personal API keys APIs

qcastel:
This PR goes in pair with https://github.com/dataiku/dip/pull/15723

You will need to checkout this PR to run the integration tests of the other PR

### DIFF
--- a/dataikuapi/dss/admin.py
+++ b/dataikuapi/dss/admin.py
@@ -1060,6 +1060,38 @@ class DSSGlobalApiKey(object):
             body = definition)
 
 
+class DSSPersonalApiKey(object):
+    """
+    A personal API key on the DSS instance
+    """
+    def __init__(self, client, id_):
+        """Do not call that directly, use :meth:`dataikuapi.DSSClient.get_personal_api_key`"""
+        self.client = client
+        self.id_ = id_
+
+    ########################################################
+    # Key description
+    ########################################################
+
+    def get_definition(self):
+        """
+        Get the API key's definition
+        :returns: the personal API key definition, as a JSON object
+        """
+        return self.client._perform_json(
+            "GET", "/personal-api-keys/%s" % (self.id_))
+
+    ########################################################
+    # Key deletion
+    ########################################################
+
+    def delete(self):
+        """
+        Delete the API key
+        """
+        return self.client._perform_empty(
+            "DELETE", "/personal-api-keys/%s" % self.id_)
+
 class DSSCluster(object):
     """
     A handle to interact with a cluster on the DSS instance
--- a/dataikuapi/dss/admin.py
+++ b/dataikuapi/dss/admin.py
@@ -1,9 +1,8 @@
 import datetime
 
-from dataikuapi.dss.utils import DSSTaggableObjectListItem
-
 from .future import DSSFuture
 import json, warnings
+
 
 class DSSConnectionInfo(dict):
     """A class holding read-only information about a connection.
@@ -1182,36 +1181,36 @@ class DSSPersonalApiKeyListItem(dict):
 
     def to_personal_api_key(self):
         """Gets the :class:`DSSPersonalApiKey` corresponding to this item"""
-        return DSSPersonalApiKey(self.client, self._data["id"])
+        return DSSPersonalApiKey(self.client, self["id"])
 
     @property
     def id(self):
-        return self._data["id"]
+        return self["id"]
    
     @property
     def user(self):
-        return self._data["user"]
+        return self["user"]
    
     @property
     def key(self):
-        return self._data["key"]
+        return self["key"]
    
     @property
     def label(self):
-        return self._data["label"]
+        return self["label"]
    
     @property
     def description(self):
-        return self._data["description"]
+        return self["description"]
    
     @property
     def created_on(self):
-        timestamp = self._data["createdOn"]
+        timestamp = self["createdOn"]
         return datetime.datetime.fromtimestamp(timestamp / 1000) if timestamp > 0 else None
    
     @property
     def created_by(self):
-        return self._data["createdBy"]
+        return self["createdBy"]
 
 
 class DSSCluster(object):
@@ -1350,7 +1349,7 @@ class DSSClusterSettings(object):
 class DSSClusterStatus(object):
     """
     The status of a cluster.
-    Do not create this object directly, use :meth:`DSSCluster.get_Status` instead.
+    Do not create this object directly, use :meth:`DSSCluster.get_status` instead.
     """
     def __init__(self, client, cluster_id, status):
         self.client = client
--- a/dataikuapi/dss/admin.py
+++ b/dataikuapi/dss/admin.py
@@ -1065,6 +1065,7 @@ class DSSGlobalApiKey(object):
 class DSSPersonalApiKey(object):
     """
     A personal API key on the DSS instance
+    Do not call that directly, use :meth:`dataikuapi.DSSClient.get_personal_api_key`
     """
     def __init__(self, client, id_):
         """Do not call that directly, use :meth:`dataikuapi.DSSClient.get_personal_api_key`"""
@@ -1078,6 +1079,7 @@ class DSSPersonalApiKey(object):
     def get_definition(self):
         """
         Get the API key's definition
+        
         :returns: the personal API key definition, as a JSON object
         """
         return self.client._perform_json(
--- a/dataikuapi/dss/admin.py
+++ b/dataikuapi/dss/admin.py
@@ -7,16 +7,15 @@ import json, warnings
 
 class DSSConnectionInfo(dict):
     """A class holding read-only information about a connection.
-    This class should not be created directly. Instead, use :meth:`DSSConnection.get_info`
+    Do not create this object directly, use :meth:`DSSConnection.get_info` instead.
 
     The main use case of this class is to retrieve the decrypted credentials for a connection,
     if allowed by the connection permissions.
 
     Depending on the connection kind, the credential may be available using :meth:`get_basic_credential` 
-    or :meth:`get_aws_credential`
+    or :meth:`get_aws_credential`.
     """
     def __init__(self, data):
-        """Do not call this directly, use :meth:`DSSConnection.get_info`"""
         super(DSSConnectionInfo, self).__init__(data)
 
     def get_type(self):
@@ -54,9 +53,9 @@ class DSSConnectionInfo(dict):
 class DSSConnection(object):
     """
     A connection on the DSS instance.
+    Do not create this object directly, use :meth:`dataikuapi.DSSClient.get_connection` instead.
     """
     def __init__(self, client, name):
-        """Do not call this directly, use :meth:`dataikuapi.DSSClient.get_connection`"""
         self.client = client
         self.name = name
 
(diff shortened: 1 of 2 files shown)


FChataigner (approved):

dataikuapi/dss/admin.py
```diff
@@ -1062,6 +1062,38 @@ def set_definition(self, definition):
...
+        self.id_ = id_
+
+    ########################################################
+    # Key description
+    ########################################################
+
+    def get_definition(self):
+        """
+        Get the API key's definition
+        :returns: the personal API key definition, as a JSON object
```
mhoarau, line 1081: Missing a new line.

mhoarau (changes requested):

dataikuapi/dssclient.py
```diff
@@ -644,6 +644,95 @@ def create_global_api_key(self, label=None, description=None, admin=False):
...
+        """
+        Get a handle to interact with a specific Personal API key
+
+        :param str key: the secret key of the desired API key
+        :returns: A :class:`dataikuapi.dss.admin.DSSPersonalApiKey` API key handle
+        """
+        return DSSPersonalApiKey(self, key)
+
+
+    def create_personal_api_key(self, label="", description="", as_type='dict'):
```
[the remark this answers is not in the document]
qcastel: That is true. Should we authorise USER_FROM_UI as well then?

dataikuapi/dssclient.py
```diff
@@ -644,6 +644,95 @@ def create_global_api_key(self, label=None, description=None, admin=False):
...
+        :returns: All personal API keys associated with your user
+        """
+        resp = self._perform_json(
+            "GET", "/personal-api-keys/")
+        if as_type == 'objects':
+            return [DSSPersonalApiKey(self, item['id']) for item in resp]
+        else:
+            return resp
+
+    def get_personal_api_key(self, key):
```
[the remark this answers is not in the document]
mhoarau: The doc now says: "the `secret` id".

FChataigner (approved):

mhoarau (changes requested):

dataikuapi/dssclient.py
```diff
@@ -644,6 +646,105 @@ def create_global_api_key(self, label=None, description=None, admin=False):
         key = resp.get('key', '')
         return DSSGlobalApiKey(self, key)
 
+    ########################################################
+    # Personal API Keys
+    ########################################################
+
+    def list_personal_api_keys(self, as_type='dict'):
```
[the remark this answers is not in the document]
qcastel: as discussed on slack, we keep dict as create_personal_api_key already existed and was returning dict by default.
For retro-compatiblity reason, we will not change the default return type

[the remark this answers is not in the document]
qcastel: actually after discussion with Fred and Mickael, it's been agreed that the new endpoints would have the default type 'object' but the create_personal_api_key will be the only one having `dict` as default for retro-compatiblity reason

FChataigner (approved):

dataikuapi/dssclient.py
```diff
@@ -667,6 +670,105 @@ def create_global_api_key(self, label=None, description=None, admin=False):
         key = resp.get('key', '')
         return DSSGlobalApiKey(self, key)
 
+    ########################################################
+    # Personal API Keys
+    ########################################################
+
+    def list_personal_api_keys(self, as_type='object'):
```
mhoarau, line 677: Careful : mismatch between `object` and `objects` (also in the doc).
So I did not test but if we follow the doc and/or use the default value you'll get a `dict`.

mhoarau (changes requested):

dataikuapi/dssclient.py
qcastel: Good catch!

dataikuapi/dssclient.py
```diff
@@ -667,6 +670,105 @@ def create_global_api_key(self, label=None, description=None, admin=False):
...
+        resp = self._perform_json(
+            "POST", "/personal-api-keys/", body={"label": label, "description": description})
+        if resp is None:
+            raise Exception('API key creation returned no data')
+        if resp.get('messages', {}).get('error', False):
+            raise Exception('API key creation failed : %s' % (json.dumps(resp.get('messages', {}).get('messages', {}))))
+        if not resp.get('id', False):
+            raise Exception('API key creation returned no key')
+
+        if as_type == 'objects':
```
mhoarau, line 723: IMO it should be plural only when a list of object is returned.

mhoarau (changes requested):

dataikuapi/dssclient.py
```diff
@@ -667,6 +670,105 @@ def create_global_api_key(self, label=None, description=None, admin=False):
...
 
+    ########################################################
+    # Personal API Keys
+    ########################################################
+
+    def list_personal_api_keys(self, as_type='objects'):
+        """
+        List all your personal API keys
+
+        :param str as_type: How to return the personal API keys. Possible values are "dict" and "object"
```
mhoarau, line 681: Still missing plural to object here (and other places)

mhoarau (changes requested):

mhoarau (approved):

dataikuapi/dss/admin.py
```diff
@@ -1169,6 +1171,38 @@ def delete(self):
         return self.client._perform_empty(
             "DELETE", "/personal-api-keys/%s" % self.id_)
 
+class DSSPersonalApiKeyListItem(DSSTaggableObjectListItem):
+    """An item in a list of datasets. Do not instantiate this class, :meth:`dataikuapi.DSSClient.list_personal_api_keys` or :meth:`dataikuapi.DSSClient.list_all_personal_api_keys`"""
```
mhoarau, line 1175: String should be on multiple lines and better formatted.

mhoarau, line 1175: it says "list of datasets".

dataikuapi/dss/admin.py
```diff
@@ -1169,6 +1171,38 @@ def delete(self):
         return self.client._perform_empty(
             "DELETE", "/personal-api-keys/%s" % self.id_)
 
+class DSSPersonalApiKeyListItem(DSSTaggableObjectListItem):
```
mhoarau, line 1174: An API key is not a `taggable` object. It does not have tags. 
FYI: `com.dataiku.dip.server.services.TaggableObjectsService.TaggableObject`
IMO: could simply extends a `dict`.
Should be

dataikuapi/dss/admin.py
```diff
@@ -1136,6 +1138,71 @@ def set_definition(self, definition):
             body = definition)
 
 
+class DSSPersonalApiKey(object):
+    """
+    A personal API key on the DSS instance
+    Do not call that directly, use :meth:`dataikuapi.DSSClient.get_personal_api_key`
```
mhoarau, line 1144: We need a coma between the two sentences.

dataikuapi/dss/admin.py
mhoarau, line 1175: We are missing a few words.
BTW we have in the 3 classes close to this one 3 different wording to say the same thing. We could at least chose one here and use it for the 3 of them:
> Do not create this object directly, use :meth:`my_method` instead.

WDYT?

dataikuapi/dssclient.py
```diff
@@ -674,21 +674,24 @@ def create_global_api_key(self, label=None, description=None, admin=False):
     # Personal API Keys
     ########################################################
 
-    def list_personal_api_keys(self, as_type='objects'):
+    def list_personal_api_keys(self, as_type='listitems'):
         """
         List all your personal API keys.
 
         :param str as_type: How to return the personal API keys. Possible values are "dict" and "objects"
```
mhoarau, line 681: Still says `dict`. (same for the `all` method)

dataikuapi/dssclient.py
```diff
@@ -674,21 +674,24 @@ def create_global_api_key(self, label=None, description=None, admin=False):
...
         
-        :return: if as_type=dict, each personal API keys is returned as a dict.
+        :return: if as_type=listitems, each one as a :class:`dataikuapi.dss.admin.DSSPersonalApiKeyListItem`.
```
mhoarau, line 683: Like the line below I would replace `one` by `key`. (same for the `all` method)

dataikuapi/dss/admin.py
```diff
@@ -1136,6 +1138,71 @@ def set_definition(self, definition):
...
+        self.client = client
+
+    def to_personal_api_key(self):
+        """Gets the :class:`DSSPersonalApiKey` corresponding to this item"""
+        return DSSPersonalApiKey(self.client, self._data["id"])
+
+    @property
+    def id(self):
+        return self._data["id"]
+    @property
```
mhoarau, line 1187: Missing blank line.

dataikuapi/dss/admin.py
```diff
@@ -1136,6 +1138,71 @@ def set_definition(self, definition):
...
+        return self._data["key"]
+    @property
+    def label(self):
+        return self._data["label"]
+    @property
+    def description(self):
+        return self._data["description"]
+    @property
+    def created_on(self):
+        return self._data["createdOn"]
```
mhoarau, line 1201: We tend to return a datetime more and more.
E.g. `dataikuapi.dss.admin.DSSUserActivity.last_successful_login`.

mhoarau (changes requested):

dataikuapi/dss/admin.py
qcastel: Good idea. I just aligned all the classes to it.

dataikuapi/dss/admin.py
```diff
@@ -1136,12 +1137,89 @@ def set_definition(self, definition):
...
+        super(DSSPersonalApiKeyListItem, self).__init__(data)
+        self.client = client
+
+    def to_personal_api_key(self):
+        """Gets the :class:`DSSPersonalApiKey` corresponding to this item"""
+        return DSSPersonalApiKey(self.client, self._data["id"])
+
+    @property
+    def id(self):
+        return self._data["id"]
```
mhoarau: Testing any properties leads to:
```py
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-825eb33a416e> in <module>()
      1 personal_keys = client.list_personal_api_keys(as_type="listitems")
      2 test = personal_keys[3]
----> 3 date = test.created_on
      4 date.isoformat()

/Users/mhoarau/devenv100/dataiku-api-client-python/dataikuapi/dss/admin.py in created_on(self)
   1207     @property
   1208     def created_on(self):
-> 1209         timestamp = self._data["createdOn"]
   1210         return datetime.datetime.fromtimestamp(timestamp / 1000) if timestamp > 0 else None
   1211 

AttributeError: 'DSSPersonalApiKeyListItem' object has no attribute '_data'
```

mhoarau (changes requested):
List item is broken (and fixed by the previous commit). Can you double check there are no other issues?

dataikuapi/dss/admin.py
qcastel: Fixed by the commit pushed by Mickael directly on the PR

PR #227, opened 2022-04-06
Add k8s cluster actions

marchelleboid:
Also adds the new "force stop" option.

[sc-82511]

### DIFF
--- a/dataikuapi/dss/admin.py
+++ b/dataikuapi/dss/admin.py
@@ -1178,7 +1178,7 @@ class DSSCluster(object):
             "POST", "/admin/clusters/%s/k8s/kubectl" % self.cluster_id,
             body={'args': args})
 
-    def delete_finished_jobs(self, delete_failed=False, namespace="", label_filter="", dry_run=False):
+    def delete_finished_jobs(self, delete_failed=False, namespace=None, label_filter=None, dry_run=False):
         """
         Runs a kubectl command to delete finished jobs.
 
@@ -1188,30 +1188,32 @@ class DSSCluster(object):
         :param str namespace: the namespace in which to delete the jobs
         :param str label_filter: delete only jobs matching a label filter
         :param bool dry_run: if True, execute the command as a "dry run"
-        :return: a dict containing the return value, standard output, and standard error of the command
+        :return: a dict containing whether the deletion succeeded, a list of deleted job names, and
+            debug info for the underlying kubectl command
         :rtype: dict
         """
         return self.client._perform_json(
             "POST", "/admin/clusters/%s/k8s/delete-finished-jobs" % self.cluster_id,
             params={'deleteFailed': delete_failed, 'namespace': namespace, 'labelFilter': label_filter, 'dryRun': dry_run})
 
-    def delete_succeeded_and_failed_pods(self, namespace="", label_filter="", dry_run=False):
+    def delete_finished_pods(self, namespace=None, label_filter=None, dry_run=False):
         """
-        Runs a kubectl command to delete succeeded and failed pods.
+        Runs a kubectl command to delete finished (succeeded and failed) pods.
 
         This operation is only valid for a Kubernetes cluster.
 
         :param str namespace: the namespace in which to delete the pods
         :param str label_filter: delete only pods matching a label filter
         :param bool dry_run: if True, execute the command as a "dry run"
-        :return: a dict containing the return value, standard output, and standard error of the command
+        :return: a dict containing whether the deletion succeeded, a list of deleted pod names, and
+            debug info for the underlying kubectl command
         :rtype: dict
         """
         return self.client._perform_json(
-            "POST", "/admin/clusters/%s/k8s/delete-succeeded-and-failed-pods" % self.cluster_id,
+            "POST", "/admin/clusters/%s/k8s/delete-finished-pods" % self.cluster_id,
             params={'namespace': namespace, 'labelFilter': label_filter, 'dryRun': dry_run})
 
-    def delete_all_pods(self, namespace="", label_filter="", dry_run=False):
+    def delete_all_pods(self, namespace=None, label_filter=None, dry_run=False):
         """
         Runs a kubectl command to delete all pods.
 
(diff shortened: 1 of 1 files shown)


alex-dataiku (approved):
LGTM

mhoarau (approved):
Did not test though.

dataikuapi/dss/admin.py
```diff
@@ -1143,22 +1143,114 @@ def start(self):
...
+        This operation is only valid for a Kubernetes cluster.
+
+        :param delete_failed: if True, delete both completed and failed jobs, otherwise only delete completed jobs
+        :type delete_failed: bool
+        :param namespace: the namespace in which to delete the jobs
+        :type namespace: str
+        :param label_filter: delete only jobs matching a label filter
+        :type label_filter: str
+        :param dry_run: if True, execute the command as a "dry run"
+        :type dry_run: bool
```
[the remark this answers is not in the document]
marchelleboid: I figured we would expose the same options, maybe it could be useful in some cases?

dataikuapi/dss/admin.py
```diff
@@ -1143,22 +1143,114 @@ def start(self):
...
+        :type args: str
+        :return: a dict containing the return value, output, and possible error output of the command
+        :rtype: dict
+        """
+        return self.client._perform_json(
+            "POST", "/admin/clusters/%s/k8s/kubectl" % self.cluster_id,
+            body={'args': args})
+
+    @staticmethod
+    def _build_args(base_command, namespace, label_filter, dry_run=True):
```
alex-dataiku, line 1185: I think this is not used anymore.

alex-dataiku (approved):
Besides the unused [method](https://github.com/dataiku/dataiku-api-client-python/pull/227/files#r849463113), LGTM (tested locally in notebooks with both admin and non admin users).

dataikuapi/dss/admin.py
```diff
@@ -1143,22 +1143,103 @@ def start(self):
...
+        :param args: the arguments to pass to kubectl (without the "kubectl")
+        :type args: str
+        :return: a dict containing the return value, output, and possible error output of the command
+        :rtype: dict
+        """
+        return self.client._perform_json(
+            "POST", "/admin/clusters/%s/k8s/kubectl" % self.cluster_id,
+            body={'args': args})
+
+    def delete_finished_jobs(self, delete_failed=False, namespace="", label_filter="", dry_run=False):
```
cstenac, line 1184: It would be preferable to have None, and to rely on None/non-None versus empty string

dataikuapi/dss/admin.py
```diff
@@ -1143,22 +1143,103 @@ def start(self):
...
 
-    def stop(self, terminate=True):
+    def stop(self, terminate=True, force_stop=False):
         """
         Stops or detaches the cluster
 
         This operation is only valid for a managed cluster.
-        :param boolean terminate: whether to delete the cluster after stopping it
+
+        :param terminate: whether to delete the cluster after stopping it
```
cstenac, line 1152: We usually put the type in the same line as the param, is there a reason to change ?

dataikuapi/dss/admin.py
```diff
@@ -1143,22 +1143,103 @@ def start(self):
...
+        """
+        Runs an arbitrary kubectl command on the cluster.
+
+        This operation is only valid for a Kubernetes cluster.
+
+        Note: this call requires an API key with DSS instance admin rights
+
+        :param args: the arguments to pass to kubectl (without the "kubectl")
+        :type args: str
+        :return: a dict containing the return value, output, and possible error output of the command
```
cstenac, line 1177: Nitpick: I would say "the return value, standard output and standard error of the command"

dataikuapi/dss/admin.py
```diff
@@ -1143,22 +1143,103 @@ def start(self):
...
+
+        :param delete_failed: if True, delete both completed and failed jobs, otherwise only delete completed jobs
+        :type delete_failed: bool
+        :param namespace: the namespace in which to delete the jobs
+        :type namespace: str
+        :param label_filter: delete only jobs matching a label filter
+        :type label_filter: str
+        :param dry_run: if True, execute the command as a "dry run"
+        :type dry_run: bool
+        :return: a dict containing the return value, output, and possible error output of the command
```
cstenac, line 1198: Shouldn't we have a more structured output here ? For this, the fact that it relies on kubectl is more of an implementation detail. Ideally, we'd return something like
```
{
   success: True/False,
   deleted: [ { "namespace": "...", "id" : ...} ],
   debug: {
       stdout:
       stderr
    }
}
```

dataikuapi/dss/admin.py
```diff
@@ -1143,22 +1143,103 @@ def start(self):
...
+        :rtype: dict
+        """
+        return self.client._perform_json(
+            "POST", "/admin/clusters/%s/k8s/delete-finished-jobs" % self.cluster_id,
+            params={'deleteFailed': delete_failed, 'namespace': namespace, 'labelFilter': label_filter, 'dryRun': dry_run})
+
+    def delete_succeeded_and_failed_pods(self, namespace="", label_filter="", dry_run=False):
```
cstenac, line 1205: so ... "finished" ? rather than "succeeded and failed "?

dataikuapi/dss/admin.py
```diff
@@ -1143,22 +1143,103 @@ def start(self):
...
+
+        This operation is only valid for a Kubernetes cluster.
+
+        :param namespace: the namespace in which to delete the pods
+        :type namespace: str
+        :param label_filter: delete only pods matching a label filter
+        :type label_filter: str
+        :param dry_run: if True, execute the command as a "dry run"
+        :type dry_run: bool
+        :return: a dict containing the return value, output, and possible error output of the command
```
cstenac, line 1217: ditto

dataikuapi/dss/admin.py
```diff
@@ -1143,22 +1143,103 @@ def start(self):
...
```
cstenac, line 1236: ditto

cstenac (changes requested):

dataikuapi/dss/admin.py
marchelleboid: Yes 😅

dataikuapi/dss/admin.py
marchelleboid: Hmm, it doesn't seem like it

dataikuapi/dss/admin.py
marchelleboid: We can add some parsing on the backend side to make the output nicer. I guess we could even use `-o name` on the delete command, but then the debug.stdout wouldn't be too interesting

mhoarau (changes requested):

mhoarau (changes requested):
LGTM, just adding "request changes" in case someone else than Marc and I has an opinion on the `dry_run` or that we decide to keep it.

mhoarau (approved):
LGTM

PR #231, opened 2022-04-20
MLflow dss flavor

vthorey:
Companion PR of https://github.com/dataiku/dip/pull/15891

* Fix and simplify the download_artifacts method
* Reset tracking_uri and end eventual run when calling `setup_mlflow`


lpenet (approved):
LGTM+Tested

PR #237, opened 2022-05-05
Add MLflow export to DSSTrainedPredictionModelDetails

vthorey:
Companion PR of https://github.com/dataiku/dip/pull/16248


lpenet (approved):
LGTM+Tested (testing that the model can be loaded and scored, either as pyfunc or dataiku flavor, with model downloaded through API or GUIà

PR #240, opened 2022-05-19
Feature/sc 81845 review notebook code samples

lpenet:
Companion API PR of https://github.com/dataiku/dku-contrib-private/pull/158 .
Took the opportunity to make minor adjustements to the public API.


vthorey (approved):
Tested  with code samples from https://github.com/dataiku/dku-contrib-private/pull/158 and LGTM

PR #246, opened 2022-07-13
[SC-93184] Create public API for project libraries

ehysi:
Add project libraries methods in project object
Adapt the plugins client to be consistent with modification in the public API

See https://github.com/dataiku/dip/pull/17144 for details on the modifications to the public API

### DIFF
--- a/dataikuapi/dss/plugin.py
+++ b/dataikuapi/dss/plugin.py
@@ -177,23 +177,23 @@ class DSSPlugin(object):
         data = f.read() # eat it all, because making it work with a path variable and a MultifilePart in swing looks complicated
         return self.client._perform_empty("POST", "/plugins/%s/contents/%s" % (self.plugin_id, path), raw_body=data)
 
-    def rename_library_file(self, path, params):
+    def rename_file(self, path, new_name):
         """
         Rename a file/folder in the plugin
 
         :param str path: the path of the file/folder, relative ot the root of the plugin
-        :param dict params: the parameters containing the new name of the file/folder
+        :param str new_name: the new name of the file/folder
         """
-        return self.client._perform_empty("POST", "/plugins/%s/contents/rename/%s" % (self.plugin_id, path), body=params)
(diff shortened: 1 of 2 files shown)


dataikuapi/dss/plugin.py
```diff
@@ -177,6 +177,23 @@ def put_file(self, path, f):
...
         return self.client._perform_empty("POST", "/plugins/%s/contents/%s" % (self.plugin_id, path), raw_body=data)
 
+    def rename_library_file(self, path, params):
+        """
+        Rename a file/folder in the plugin
+
+        :param str path: the path of the file/folder, relative ot the root of the plugin
+        :param dict params: the parameters containing the new name of the file/folder
+        """
+        return self.client._perform_empty("POST", "/plugins/%s/contents/rename/%s" % (self.plugin_id, path), body=params)
```
StanislasBertrand, line 187: Since the required structure of the params object is not obvious to someone working only with the API, we could either 
* Specify in the method description that params should have a "newName" key
* Make the method take directly the name as an argument and build the `params = {"newName": name}` in the method

Same goes for the other methods in libraries & plugins apis.
What do you think ?

StanislasBertrand (changes requested):

dataikuapi/dss/plugin.py
```diff
@@ -177,6 +177,23 @@ def put_file(self, path, f):
         data = f.read() # eat it all, because making it work with a path variable and a MultifilePart in swing looks complicated
         return self.client._perform_empty("POST", "/plugins/%s/contents/%s" % (self.plugin_id, path), raw_body=data)
 
+    def rename_library_file(self, path, params):
```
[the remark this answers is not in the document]
ehysi: copy/paste issue

dataikuapi/dss/plugin.py
ehysi: I think the second one is the better because more explicit and less prone to errors from a client perspective

StanislasBertrand (approved):
LGTM

PR #249, opened 2022-07-26
Feature/dss110 sc 94866 add a compute prediction drift function to

pmmayne:


lpenet:
I wonder if having a more general "compute drift" method, computing both drifts, maybe optionally, would be better?

pmmayne:
> I wonder if having a more general "compute drift" method, computing both drifts, maybe optionally, would be better?

I can add this "compute drift" general method which will call both compute_data_drift and compute_prediction_drift, I think it is a good idea

PR #251, opened 2022-07-29
Give possibility to API user to override the external model evaluate …

pmmayne:
…sampling parameter

Companion PR: https://github.com/dataiku/dip/pull/17348

### DIFF
--- a/dataikuapi/dss/savedmodel.py
+++ b/dataikuapi/dss/savedmodel.py
@@ -1,5 +1,6 @@
 import tempfile
 
+from dataikuapi.dss.utils import DSSDatasetSelectionBuilder
 from .metrics import ComputedMetrics
 from .ml import DSSMLTask
 from .ml import DSSTrainedClusteringModelDetails
@@ -354,7 +355,7 @@ class MLFlowVersionHandler:
             "/projects/%s/savedmodels/%s/versions/%s/external-ml/metadata" % (self.saved_model.project_key, self.saved_model.sm_id, self.version_id),
             body=metadata)
 
-    def evaluate(self, dataset_ref, container_exec_config_name="INHERIT"):
+    def evaluate(self, dataset_ref, container_exec_config_name="INHERIT", override_selection=None):
         """
         Evaluates the performance of this model version on a particular dataset.
         After calling this, the "result screens" of the MLFlow model version will be available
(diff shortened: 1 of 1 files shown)


dataikuapi/dss/savedmodel.py
```diff
@@ -366,12 +367,18 @@ def evaluate(self, dataset_ref, container_exec_config_name="INHERIT"):
         :param str container_exec_config_name: Name of the containerized execution configuration to use for running the evaluation process.
                                   If value is "INHERIT", the container execution configuration of the project will be used.
                                   If value is "NONE", local execution will be used (no container)
+        :param str override_selection: will default to HEAD_SEQUENTIAL with a maxRecords of 10_000.
+        :type card: :class:`DSSDatasetSelectionBuilder` or dict (obtained from ``get_raw_sampling_selection()``)
```
fcandela-dku, line 371: I am not able to find the method `get_raw_sampling_selection`. Where is it?

pmmayne: I don't know if this methods exists yet or no, I just assumed it would be best to mimic the doc string of `dataikuapi/dss/statistics.py`:
```
    def set_sampling_settings(self, selection):
        """
        Sets the sampling settings of the worksheet

        :type card: :class:`DSSDatasetSelectionBuilder` or dict (obtained from ``get_raw_sampling_selection()``)
        """
        raw_selection = selection.build() if isinstance(
            selection, DSSDatasetSelectionBuilder) else selection
        self._worksheet_definition['dataSpec']['datasetSelection'] = raw_selection
```

I have also looked for the method but it seems it does not exist (yet? anymore?). I can remove the part about the unknown method and keep only:
```
 :type card: :class:`DSSDatasetSelectionBuilder` or dict (obtained from ``get_raw_sampling_selection()``)
```

fcandela-dku: I think it makes reference to the method `get_raw_sampling_settings()`, so maybe it changed the named and it was forgotten to update the docs. So I agree to keep just the option of providing a `DSSDatasetSelectionBuilder` or `None` but not a dict, since in our case we have no other alternatives.
I have also noticed that the comment says :type card: :class:`DSSDatasetSelectionBuilder` instead of :type override_selection: :class:`DSSDatasetSelectionBuilder`.

fcandela-dku (changes requested):
See required changes in PR comments

fcandela-dku (approved):
LGTM+Tested!

PR #254, opened 2022-08-05
dataset import: list and import elastic search indices

jfyuen:
Add listing and importing ES indices via the public API.
Name spacing and the "tables" are kept for backward compatibility and to prevent adding another object (seen with @cstenac).

[sc-95861]


gssiyankai (approved):
LGTM

PR #259, opened 2022-08-25
Add wrapper for workspaces

adescamps:
Add wrapper for workspace API, all actions on the API is supported.

Not sure if we want to share to a workspace through a DSSDataset, DSSArticle, etc so it's not done yet but can be added if needed.

[sc-97871]

### DIFF
--- a/dataikuapi/dss/workspace.py
+++ b/dataikuapi/dss/workspace.py
@@ -1,0 +1,101 @@
+from dataikuapi.dss.app import DSSApp
+from dataikuapi.dss.dataset import DSSDataset
+from dataikuapi.dss.wiki import DSSWikiArticle
+
+
+class DSSWorkspace:
+    """
+    A handle to interact with a workspace on the DSS instance.
+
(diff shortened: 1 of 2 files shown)


dataikuapi/dss/workspace.py
```diff
@@ -0,0 +1,101 @@
...
+            data = {"reference": {"projectKey": object.project_key, "type": "DATASET", "id": object.dataset_name}}
+        elif isinstance(object, DSSWikiArticle):
+            data = {"reference": {"projectKey": object.project_key, "type": "ARTICLE", "id": object.article_id}}
+        elif isinstance(object, DSSApp):
+            data = {"appId": object.app_id}
+        elif isinstance(object, DSSWorkspaceHtmlLinkObject):
+            data = {"htmlLink": {"name": object.name, "url": object.url, "description": object.description}}
+        elif isinstance(object, dict):
+            data = object
+        else:
```
adescamps, line 50: We are missing a case for DSSDashboard but the class is not yet available as the API is not yet merged.
I havent seen a handle for a webapp so the user will need to use the dict form

PR #265, opened 2022-09-08
99387 : support creating a govern node

ericcournarie:
add support for govern node creation

`
client = dataikuapi.FMClientAWS ("https://somewhere", "key", "secret")
client._session.verify = None

creator = client.new_virtual_network_creator ("MyNetwork")
creator.with_vpc ("vpc-xxx", "subnet-yyy")
netw = creator.create ()
netw.set_fleet_management (True, govern_server="Governor")
netw.save()
`
`
creator = client.new_instance_creator("MyGovernInstance", "ist-IJcznImxJFTI", "vn-xWfdn88cADov", "dss-11.0.2-default")
creator.with_dss_node_type(dataikuapi.fm.instances.FMNodeType.GOVERN).with_cloud_instance_type("m5a.2xlarge")
dss = creator.create ()
`

### DIFF
--- a/dataikuapi/fm/instances.py
+++ b/dataikuapi/fm/instances.py
@@ -37,14 +37,12 @@ class FMInstanceCreator(object):
         """
         Set the DSS Node type of the instance to create
 
-        :param str dss_node_type: Optional , the type of the dss node to create. Supports "design", "automation ordeployer". Defaults to "design"
+        :param str dss_node_type: Optional , the type of the dss node to create. Defaults to `DESIGN`.
+        :type dss_nodetype: :class:`dataikuapi.fm.instances.FMNodeType`
         :rtype: :class:`dataikuapi.fm.instances.FMInstanceCreator`
         """
-        if dss_node_type not in ["design", "automation", "deployer"]:
-            raise ValueError(
-                'Only "design", "automation" or "deployer" dss_node_type are supported'
-            )
-        self.data["dssNodeType"] = dss_node_type
+
+        self.data["dssNodeType"] = dss_node_type.value
         return self
 
     def with_cloud_instance_type(self, cloud_instance_type):
@@ -389,6 +387,11 @@ class FMGCPInstance(FMInstance):
         self.instance_data["gcpPublicIPId"] = public_ip_id
         return self
 
+class FMNodeType(Enum):
+    DESIGN = "design"
+    DEPLOYER = "deployer"
+    AUTOMATION = "automation"
+    GOVERN = "govern"
 
 class FMInstanceEncryptionMode(Enum):
     NONE = "NONE"
--- a/dataikuapi/fm/instances.py
+++ b/dataikuapi/fm/instances.py
@@ -15,7 +15,7 @@ class FMInstanceCreator(object):
         self, client, label, instance_settings_template_id, virtual_network_id, image_id
     ):
         """
-        Helper to create a DSS Instance
+        Helper to create a DSS instance.
 
         :param object client: :class:`dataikuapi.fm.fmclient`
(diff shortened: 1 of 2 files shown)


dataikuapi/fm/instances.py
```diff
@@ -37,14 +37,12 @@ def with_dss_node_type(self, dss_node_type):
         """
         Set the DSS Node type of the instance to create
 
-        :param str dss_node_type: Optional , the type of the dss node to create. Supports "design", "automation ordeployer". Defaults to "design"
+        :param dss_node_type: Optional , the type of the dss node to create. Defaults to `DESIGN`.
```
spicquenot, line 40: I don't know if we want to be compatible with old usage, but with this change it will break (AttributeError: 'str' object has no attribute 'value')

If we do not want to break, either:
- cover the two types with a check: isinstance(dss_node_type, FMNodeType) 
- just add a string govern instead
- have the string value builder outside of this method

Quick note: maybe `def with_dss_node_type(self, dss_node_type=FMNodeType.DESIGN):` in the signature to make it "really" optional as a technical method or dss_node_type="design" for string

spicquenot (commented):
I don't know enough FM to verify the code that it will work.

However I have put a small remark about the compat with older version

dataikuapi/fm/instances.py
ericcournarie: > Quick note: maybe `def with_dss_node_type(self, dss_node_type=FMNodeType.DESIGN):` in the signature to make it "really" optional as a technical method or dss_node_type="design" for string

the optional is not welcome in the doc , as the parameter is not optional .. it's the method call that is optional, as it has a default value of DESIGN if you do not use `with_dss_node_type` . there is another story about cleaning comments though ..

ericcournarie: > I don't know if we want to be compatible with old usage

not sure it is needed, as I doubt customers knows there was an API before .. maybe it's better to have clean script now . my 2 cents .. Fred will tell us its thoughts hopefully :-)

spicquenot: Ah yes, indeed, in this case, I would put the defaults to design only in the other method that actually put the default to design and I would remove the word optional from here in the doc. Other it may be misleading for this particular method usage.

Noted, and ok pour waiting for the compatibility

dataikuapi/fm/instances.py
```diff
@@ -37,14 +37,12 @@ def with_dss_node_type(self, dss_node_type):
...
+        :type dss_node_type: :class:`dataikuapi.fm.instances.FMNodeType`
         :rtype: :class:`dataikuapi.fm.instances.FMInstanceCreator`
         """
-        if dss_node_type not in ["design", "automation", "deployer"]:
-            raise ValueError(
-                'Only "design", "automation" or "deployer" dss_node_type are supported'
-            )
-        self.data["dssNodeType"] = dss_node_type
+
+        self.data["dssNodeType"] = dss_node_type.value
```
FChataigner, line 45: to complement the comment from @spicquenot , I'd maintain backwards compatibility by doing something like

```
if isinstance(dss_node_type, enum):
    self.data["dssNodeType"] = dss_node_type.value
else:
    self.data["dssNodeType"] = dss_node_type
```

(or something that is python2 AND python3 compatible)

the signature and comment change is fine IMHO

FChataigner (changes requested):

dataikuapi/fm/instances.py
ericcournarie: BW compatibility added

FChataigner (approved):

PR #268, opened 2022-09-19
99387 support Python2

ericcournarie:
support python2, which does not like Enum

### DIFF
--- a/dataikuapi/fm/instances.py
+++ b/dataikuapi/fm/instances.py
@@ -43,10 +43,15 @@ class FMInstanceCreator(object):
         :rtype: :class:`dataikuapi.fm.instances.FMInstanceCreator`
         """
         # backward compatibility, was a string before . be sure the value falls into the enum
-        value = dss_node_type;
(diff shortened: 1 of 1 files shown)


dataikuapi/fm/instances.py
```diff
@@ -406,6 +411,14 @@ class FMNodeType(Enum):
     AUTOMATION = "automation"
     GOVERN = "govern"
 
+# Python2 emulated enum. to be removed on Python2 support removal
+def get_from_string(s):
```
amandineslx, line 415: should we define this method as a `@staticmethod` of class `FMNodeType`?
here, it is at the same level as the FM classes related to DSS instances in this file

amandineslx (changes requested):
not really requesting the change but requesting to consider the question ;)

dataikuapi/fm/instances.py
ericcournarie: indeed, it would be nicer to hide this method

PR #269, opened 2022-09-19
Add target variable as optional param of DSSMLTask#guess

AgatheG:
Adds an optional `target_variable` string param for` DSSMLTask#guess`.
See https://github.com/dataiku/dip/pull/17631 for Java implementation of the public API.

[sc-81879]


mhham (commented):
Wondering whether previous behaviour is preserved if target_variable is not provided.

mhham:
Should be moved to dip

PR #273, opened 2022-12-13
Expose role kwarg within SingleOutputRecipeCreator.with_output() and SingleOutputRecipeCreator.with_existing_output()

jonbannister:


jonbannister (commented):
This will prevent users of the API from having to use the private method to provide a role for the output.

Issue #272, opened 2022-12-02
Pinned requests dependency


chrisk314:
Please resolve whatever issues are preventing you from using the most recent version of `requests` and unpin this dependency. Pinned dependencies in library packages introduce issues for users downstream, as shown in the output from `pip` included below.

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
dataiku-api-client 11.0.0 requires requests<2.22,>=2, but you have requests 2.28.1 which is incompatible.
Successfully installed adlfs-2022.11.2 aiohttp-3.8.3 art-5.8 azure-storage-blob-12.14.1 backoff-2.2.1 contourpy-1.0.6 cycler-0.11.0 exceptiongroup-1.0.4 fonttools-4.38.0 importlib-metadata-5.1.0 ipywidgets-8.0.2 jsonschema-4.17.3 jupyter-1.0.0 jupyter-console-6.4.4 jupyterlab-widgets-3.0.3 kiwisolver-1.4.4 llvmlite-0.39.1 markdown-3.4.1 matplotlib-3.6.2 mlflow-2.0.1 nest_asyncio-1.5.6 networkx-2.8.8 numba-0.56.4 numpy-1.22.4 pandas-1.5.2 papermill-2.4.0 pika-1.3.1 pyarrow-10.0.1 pydantic-1.10.2 pytest-7.2.0 pytest-asyncio-0.20.2 qtconsole-5.4.0 qtpy-2.3.0 redis-4.3.5 requests-2.28.1 scikit-learn-1.1.3 shap-0.41.0 slicer-0.0.7 treelite-3.0.1 treelite_runtime-3.0.1 widgetsnbextension-4.0.3
```

janhurst:
I have the same issue here. I've cloned this repo and unpinned the dependency and built a private pip package to work around this. It has work across a number of projects.

PR #274, opened 2023-03-06
Relax requests version spec

janhurst:
Pinning requests to <2.22 causes many packages to conflict.

See #272


janhurst:
I think this has effectively been implemented already

janhurst:
[This tag ](https://github.com/dataiku/dataiku-api-client-python/blob/12.1.0/setup.py) in the repo seems to be setting the requests dependency to `requests<3` which is at least an improvement from the old pin to to `requests<2.22`

PR #290, opened 2024-02-26
Task/dss 12.5.2 sync sources

kevanescence:


kevanescence:
[sc-173550]

rpainblanc (approved):
LGTM

Checked that:
- content is identical with that of dip@12.5.2

PR #293, opened 2024-05-31
12.6.3 packaging

kevanescence:


kevanescence:
[sc-186758]

acoquard (approved):
checked diff is actually empty

PR #295, opened 2024-06-25
Task/dss 13.0.0 publish on pypi

kevanescence:


kevanescence:
[sc-190595]

rpainblanc (approved):
I checked the diff with dip@13.0.0 and found no difference with the PR, all changes there are here too
I checked pip install
All good

PR #296, opened 2024-06-28
SQL notebooks API

vojtech2:


dataikuapi/dss/sqlnotebook.py
```diff
@@ -0,0 +1,206 @@
...
+
+        Do not instantiate directly, use :meth:`dataikuapi.dss.sqlnotebook.DSSSQLNotebook.get_history`
+    """
+    def __init__(self, client, project_key, notebook_id, query_runs):
+        self.client = client
+        self.project_key = project_key
+        self.notebook_id = notebook_id
+        self.query_runs = query_runs
+
+    def get_query_runs(self):
```
vojtech2, line 184: I'm deliberately calling these query runs instead of queries. LMK if that's alright or we should use a different name.

apichery (commented):
Not tested but looks really good already :)

vojtech2:
I've now copied the changes into my dip PR. Should've made the changes there in the first place. Will close this one.

PR #297, opened 2024-07-24
Task/dss 13.0.1 publish on pypi

kevanescence:


rpainblanc (approved):
Checked the diff with dip@13.0.1 and found no difference with the PR, all changes (only a minor one) there are here too
I checked pip install
All good

PR #298, opened 2024-07-25
Task/dss 13.0.2 publish on pypi

kevanescence:


kevanescence:
[sc-195945]

rpainblanc (approved):
- Checked the diff with dip@13.0.2 and found no difference with the PR, all changes there are here too
- Checked pip install
- Checked using the API to query DSS

All good ✅

PR #299, opened 2024-08-01
13.0.3 release metadata

kevanescence:


rpainblanc (approved):
Checked the diff with dip@13.0.3 and found no difference with the PR

PR #305, opened 2024-09-19
13.1.4 Release metadata

kevanescence:


kevanescence:
[sc-203761]

rpainblanc (approved):
- Checked the diff with dip@13.1.4 and found no difference with the PR, all changes there are here too
- Checked pip install
- Checked using the API to query DSS

✅ all good

PR #311, opened 2024-11-25
Publish 13.2.3 on PyPI

kevanescence:


kevanescence:
[sc-215913]

rpainblanc (approved):
- Checked diff with dip@13.2.3 and found no difference
- Checked pip install
- Called simple API on DSS 13.2.3 and DSS 13.0.3

all good ✅ 

LGTM

PR #326, opened 2025-05-15
Publish 13.5.3 on PyPI

kevanescence:
[sc-244058]


rpainblanc (approved):
- Checked with DIP@13.5.3 and found no difference
- Checked pip install
- Checked sending API calls to DSS 13.5.2 and 13.1.x

All good ✅

PR #327, opened 2025-05-23
13.5.4 - Publish on PyPI

kevanescence:


kevanescence:
[sc-245985]

rpainblanc (approved):
- Checked with DIP@13.5.4 and found no difference
- Checked pip install
- Checked sending API calls to DSS 13.5.2 and 13.1.x

All good ✅

PR #329, opened 2025-07-16
Publish 13.5.6 on PyPI

kevanescence:


rpainblanc (approved):
- Checked with DIP@13.5.6 and found no difference
- Checked pip install
- Checked sending API calls to DSS 14.0.0 and 13.1.x
All good ✅

PR #356, opened 2026-05-25
14.5.2 publish on PyPI

kevanescence:
[sc-319016]


rpainblanc (approved):
- Checked with DIP@14.5.2 and found no difference
- Checked pip install
- Checked sending API calls to DSS 14.5.x

All good ✅

PR #365, opened 2026-08-14
15.0.0 publish on pypi

kevanescence:


kevanescence:
[sc-336890]

rpainblanc (approved):
- Checked with DIP@15.0.0 and found no difference
- Checked pip install
- Checked sending API calls to DSS 14.7.x and 15.0.0

All good ✅

PR #367, opened 2026-09-15
15.0.1 publish on PyPI

kevanescence:


kevanescence:
[sc-342794]

rpainblanc (approved):
- Checked with DIP@15.0.1 and found no difference
- Checked pip install

All good ✅

